PR #1, opened 2020-09-29
Adding section HTML/CSS and the freeCodeCamp link

MarianaSouza:
In this PR, a new HTML/CSS section was added as well as the freeCodeCamp link with the recommended course description.


rebecatschiedel (approved):
Free code camp is a great resource to start learning css and html.

PR #8, opened 2020-10-06
Adding the section books

rebecatschiedel:
Add a new section called Books, section-book/CommandLine, section-books/JavaScript(JS) and section-books/Programming.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,20 @@ This is a video that explains a little bit how to start collaborating with the p
 <br />
 <br />
 
+## Books ##
+
+* **Command Line** 
+    * [Learn Enough Command Line to Be Dangerous - Michael Hartl](https://www.learnenough.com/command-line-tutorial/basics)
+    
+* **JavaScript (JS)**
+    * [JavaScript for Impatient Programmers - Dr. Axel Raushmayer](https://exploringjs.com/impatient-js/index.html)    
+
+* **Programming**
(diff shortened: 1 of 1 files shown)


MarianaSouza:
Hey @rebecatschiedel ,

That's a great idea to add a "Books" section. Thanks!
We should try following the branch guidelines we see in the introduction section, which means your branch should be called something like section-course/books, for example. It's pretty common when you're an in a team, having to follow naming guidelines. Could you please rename the branch? https://linuxize.com/post/how-to-rename-local-and-remote-git-branch/

Also, the PR title should present the general idea of the activity you are doing here, such as "Adding the section books"

MarianaSouza (changes requested):
Hey @rebecatschiedel ,

That's a great idea to add a "Books" section. Thanks!
We should try following the branch guidelines we see in the introduction section, which means your branch should be called something like section-course/books, for example. It's pretty common when you're an in a team, having to follow naming guidelines. Could you please rename the branch? https://linuxize.com/post/how-to-rename-local-and-remote-git-branch/

Also, the PR title should present the general idea of the activity you are doing here, such as "Adding the section books"

rebecatschiedel:
Hi @MarianaSouza,

Thank you for the explanation and for the link, both were very useful!!

rebecatschiedel:
I opened a new pull request because I messed up my forked code while i was trying to rename it. It was a great opportunity to learn a lot of new things.

PR #65, opened 2021-12-20
Add Image component

panoramine:
| [**64**](https://github.com/MarianaSouza/web-dev-path/issues/64) |

#### What is this change?
Substitution of the img element for the new [Image Component](https://nextjs.org/docs/basic-features/image-optimization) for better performance and automatic Image Optimization. This change was made for the logo (Navigation bar and Footer)  and social media icons on the Footer.
The Image Component mandatory properties:
- height
- width
- src

The social media icons lost their style after the change. Based on [research](https://github.com/vercel/next.js/discussions/22861) I made and with the assistance of fellow developer @rebecatschiedel I fixed the bug adding a new section tag to wrap the Image component and then styling the section instead of trying to style the Image directly. 
Although the issue was correctly fixed, we could not understand why we cannot use padding and margins while styling the Image component. 
It was also noticed that the use of the Image component created a new DOM structure inside the section tag like this:
```
<section>
    <div>
        <div>
            <img />
        </div>
        <!-- image that we want to render -->
        <img />   
    </div>
</section>
```

This [situation](https://github.com/vercel/next.js/discussions/18820) was also noticed by other people.

Locally on my branch feature/add-image-component.

After code review.

### DIFF
--- a/components/Footer.js
+++ b/components/Footer.js
@@ -2,6 +2,7 @@ import Link from "next/link";
 import footerStyles from "../styles/Footer.module.scss";
 import { linksNav, linksSocial } from "../utils/links";
 import NewsletterSubscribe from "./mailchimp/NewsletterSubscribe";
+import Image from "next/image";
 
 export default function Footer() {
   return (
@@ -19,7 +20,7 @@ export default function Footer() {
               </li>
             ))}
           </ul>
-          <img
+          <Image width={77} height={72}
             className={footerStyles.logo}
             src="/images/web-dev-path-logo-small.png"
             alt="Logo"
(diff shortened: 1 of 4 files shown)


components/Footer.js
```diff
@@ -36,13 +37,14 @@ export default function Footer() {
...
           <div className={footerStyles.socialMedia}>
             {linksSocial.map((link) => (
-              <Link href={link.href} key={link.text}>
-                <img
-                  className={footerStyles.socialMedia}
-                  src={link.src}
-                  alt={link.text}
-                />
-              </Link>
+              <section className={footerStyles.socialMedia}>
```
MarianaSouza, line 40: Instead of wrapping the new `<Image />` component into a new element with the same `className` of its parent, which seems a little repetitive, could you please try the second approach described[ in your research](https://github.com/vercel/next.js/discussions/22861), @panoramine , please?
That means you're just adding a new class to the <Image /> component itself and, then, styling it (add a margin-right between the images, if necessary).

Screenshot 2021-12-19 173727

MarianaSouza (changes requested):
Please let me know if my comment above about the social media links makes sense, @panoramine :) Nice work so far!

MarianaSouza (approved):
I've done a commit to solve that styling issue without the section wrapper @panoramine  and @rebecatschiedel .
 If you agree and understand this code, please ask Rebeca to review it too, test it locally (or by using the Vercel PR preview), and then it's all yours to merge 🥳 
Way to go on your first PR! 🚀

rebecatschiedel (approved):
It was great to learn with you @panoramine and @MarianaSouza. It is so nice to see that we can achieve the same result in different ways and how to choose the best one for the project.

PR #72, opened 2022-02-07
Update Next.js from 10.0.0 to 12.0.10

mariana-caldas:

#### What is this change?

Next.js as well as its dependencies are being updated so that we can start developing the new Homepage from the most recent framework version.  It's changing from the version `10.0.0` to `12.0.10`.

When running some tests on the applications, I've realized the project Next.js version was way behind, which can make the codebase vulnerable. So, after running `npm install`  to check the possible necessary vulnerabilities, I run `npm audit fix --force` which updated everything.

No.

Locally and also checked the deployment preview provided by Netlify.

After review.


panoramine (commented):
Good preventative update.

PR #75, opened 2022-02-09
Feature/newsletter footer styling

manecipoglu:
| [**73**](https://github.com/MarianaSouza/web-dev-path/issues/73) |

#### What is this change?
Footer section along with the newsletter part is redesigned to match the mockup. 
Logo is extracted as SVG and added to our images.
New mixins are added for responsive layout. 
New colors are added to the variables. 
Font-family is integrated from google fonts.
Global styles are updated.
Switched to using @use for scss imports for the components modified by this issue.
Semantic html tags are used for the footer.
Added 'Join us' section to the available links in our utils.js.

Extracting the logo as SVG from the Figma file did not go as expected. 
It is literally a PNG file dressed up as SVG. Because of this and the whitespace it came around with,
styling is not implemented professionally. It looks OK but it is patched. Should be redone after we have a 
proper SVG logo. I spent a lot of time trying to get it done to no avail. 

Moreover, I tried incorporating BEM naming convention because of it's near perfect integration with 
component based design but JSS highly limits the proper use. Might be a good idea to follow another naming convention. 

I may have used poor naming for the added colors. We should take a look at it and maybe change it now before we use the
same name in other components. 

Since we have only a large desktop mockup, I improvised the layout between mobile and large desktop. 
Open to suggestions and reviews regarding that part as well. 

Locally.

After code review.


mariana-caldas:
This task has so many caveats and you are getting there like a pro @manecipoglu ! Good job! The responsiveness is great and you found a very elegant solution with the mixins for media queries 🚀 _Making a mental note here to go to the other components to update those media queries with the new variables in the next PR's 🥇_ 
 
1- When facing problems with assets from design in a team you can count on your fellow designer, don't spend a lot of time on that.  Reach out to @rebecatschiedel and @jomaira-rgb and talk about the SVG situation, please. **It seems the SVG files need to be revisited/redone on the homepage, ladies, and not only for the logo**. **Also, it's not clear on the mockup how the "hover" effect on items will be**.  Could you please take a look at that this week and let Muammer know? Thanks!

---

2- Besides the logo, we need to work a little bit more with the Newsletter error/success message for desktop and mobile, @manecipoglu . We should position it below the inputs. Also, maybe all the form items should be "flex-wrapped" on desktop.   Please reach out if you need more instructions about that.

message-form

 ---

3- Yes, we can revisit the variable color naming and add them like that, @manecipoglu :

//Colors
$primary-content-color: #292929;
$primary-bg-color: #eaeaea;
$dark-bg-color: #023047;
$light-bg-color: #8CD5E8;
$primary-accent-color: #FFCC4C;
$second-accent-color: #19AAD1;
$black: #000000;
$white: #ffffff;
$transparent: transparent;

colors

mariana-caldas:
There are just a few things we need to finish this PR, @manecipoglu ! We're getting there. Yep, the details matter, right? 

1 - We need to replace the logo PNG with the SVG I think @rebecatschiedel is finishing for us. That's why I'm keeping this PR as a work in progress label for now.

2- Could you please wrap the footer logo into a <Link /> and link it to the Homepage? 

3- Could you apply the same hover effects (opacity 0.6 with the default transition) we see on the "Learn More" button (the new section added by Shunya) **to the logo link and menu items**?

4- On mobile ( less than 768px), could you please:
 a) adjust the logo so it has a max-width of 12.5rem? You may have to use `!important` besides adding a specific class?
 b) center the menu items?
 
 It will look something like this:
mobile-footer

 Please let me know if you have any questions about it! Thanks 🚀

manecipoglu:
Thanks for the feedback Mariana! 
I think, I managed to cover the issues you pointed out. 

It's interesting how Next renders two spans for images. I want to dive deeper on that and figure out the reasoning behind.
I also wonder if we need to use anchor tags inside Link components. The documentation puts them there but avoiding them also renders anchor tags in the DOM tree. 

I moved the hover effect to the global styles since we also have the transition applied there.

mariana-caldas:
> Thanks for the feedback Mariana! I think I managed to cover the issues you pointed out.
> 
> It's interesting how Next renders two spans for images. I want to dive deeper on that and figure out the reasoning behind. I also wonder if we need to use anchor tags inside Link components. The documentation puts them there but avoiding them also renders anchor tags in the DOM tree.
> 
> I moved the hover effect to the global styles since we also have the transition applied there.

Thanks, @manecipoglu !
You don't have to use an `a` if you don't have any children. In that case, you had to because we wanted to wrap an image. I think this post can help us to understand that "interesting" 🙄  behaviour: https://stackoverflow.com/questions/67911391/nextjs-link-isnt-rendering-an-anchor-tag?noredirect=1&lq=1

mariana-caldas (approved):
LGTM! 🚀 
Thanks so much, @manecipoglu ! All yours to merge! 🎉 
It'd be great if you could review this PR too @AMSdesigns since I've changed some details about the TwoColumn component.
Also, this is a great PR to learn from @rebecatschiedel and @panoramine !

mariana-caldas (dismissed):

mariana-caldas (approved):

jomaira-rgb:
oh.. that is looking good!

PR #76, opened 2022-02-09
added TwoColumn and Button component

AMSdesigns:
| [**74**](https://github.com/MarianaSouza/web-dev-path/issues/74) |

#### What is this change?
 - Created 'TwoColumn.js' component that can receive props.
 - Created 'TwoColumn.module.scss' for styling.
 - Created 'Button.js' component that can receive props.
 - Created 'Button.module.scss' for styling.

I wasn't quite sure what to do for the image. I here didn't upload it from Figma this time yet.
As written in issues description, I tried to take Button into component too. 
For the Button component, it can take props for styling and I wrote css in 'Button.module.scss', but I guess it might not be the best way to reuse them in the future..
Please check.

Locally.

After code review.

### DIFF
--- a/components/Button.js
+++ b/components/Button.js
@@ -1,0 +1,10 @@
+export default function Button(props) {
+  const { className, text, link } = props;
+  return (
+    <>
+      <a href={link} className={className}>
+        {text}
+      </a>
+    </>
+  );
+}
--- a/components/TwoColumn.js
+++ b/components/TwoColumn.js
@@ -1,0 +1,25 @@
+import styles from '../styles/TwoColumn.module.scss';
+import buttonStyles from '../styles/Button.module.scss';
+import Button from './Button';
+
+export default function TwoColumn(props) {
+  const { image, title, content, rowOrder } = props;
+  return (
+    <div className={styles.container}>
+      <div className={styles.inner} style={{ flexDirection: rowOrder }}>
+        <div className={styles.itemLeft}>
+          <h2 className={styles.title}>{title}</h2>
+          <p className={styles.content}>{content}</p>
+          <Button
+            text="Learn More"
+            link="/about-us"
+            className={buttonStyles.btn}
+          />
+        </div>
+        <div className={styles.itemRight}>
+          <img src={image} alt="" className={styles.img} />
+        </div>
+      </div>
+    </div>
+  );
+}
--- a/styles/Button.module.scss
+++ b/styles/Button.module.scss
@@ -1,0 +1,22 @@
+@import 'variables';
+
+.btn {
+  padding: 8px 32px;
+  color: #fff;
+  background-color: #292929;
+  border-radius: 32px;
+  transition: 0.3s ease;
+  cursor: pointer;
+  font-family: 'Assistant';
+  font-style: normal;
+  font-weight: bold;
+  font-size: 24px;
+  line-height: 31px;
+  align-items: center;
+  text-align: center;
+  min-width: 256px;
+}
+
+.btn:hover {
+  opacity: 0.6;
+}
--- a/styles/TwoColumn.module.scss
+++ b/styles/TwoColumn.module.scss
@@ -1,0 +1,64 @@
+@import 'variables';
+
+.container {
+  width: 100vw;
+  margin: 0 calc(50% - 50vw);
+}
+.inner {
+  padding: 80px 0;
+  min-height: 50vh;
+  margin: 0 auto;
+  width: 90%;
+  display: flex;
+  justify-content: space-between;
+  max-width: 1440px;
+}
+
+.title {
+  font-family: 'Open Sans';
+  font-style: normal;
+  font-weight: bold;
+  font-size: 72px;
+  line-height: 80px;
+  color: #292929;
+  margin: 0;
+}
+.itemLeft {
+  width: 50%;
+  text-align: left;
+}
+.itemRight {
+  width: 50%;
+  text-align: end;
+}
+
+.content {
+  font-family: 'Assistant';
+  font-style: normal;
+  font-weight: normal;
+  font-size: 24px;
+  line-height: 31px;
+  display: flex;
+  align-items: center;
+  margin: 62px 0 40px;
+}
+
+.img {
+  object-fit: cover;
+  height: 100%;
+  width: 100%;
+  max-width: 409px;
+  max-height: 545px;
+}
+
+@media (max-width: $tablet-breakpoint) {
+  .inner {
+    flex-direction: column;
+    gap: 60px;
+  }
+  .itemLeft,
+  .itemRight {
+    width: 100%;
+    text-align: center;
+  }
+}


mariana-caldas:
Hey @AMSdesigns ! :) Nice to meet you and welcome to our project!

Your PR seems to be very well structured, I guess we need to add that component you've just created into the Homepage, right? Before that, let's talk a little bit on our Slack channel so that you know the other devs better since the Web Dev Path purpose is to learn how to work as a team. Can I have your email to send you an invite to Slack and to the repo itself? Thanks :)

AMSdesigns:
Hi @mariana-caldas, thank you for your reply.
Yes I just joined your Slack Channel, please check!

components/Button.js
```diff
@@ -0,0 +1,10 @@
+export default function Button(props) {
+  const { className, text, link } = props;
+  return (
+    <>
```
manecipoglu, line 4: Since you are returning a single element here (in this case an anchor tag (<a>) you do not need to wrap the returned jsx in react fragments (<>).

components/Button.js
```diff
@@ -0,0 +1,10 @@
...
+      <a href={link} className={className}>
```
manecipoglu, line 5: Because we are using NextJs, might be a better idea to use Link component from the framework instead of plain anchor tag.

components/Button.js
```diff
@@ -0,0 +1,10 @@
...
+        {text}
```
manecipoglu, line 6: Rather than limiting our Button component with a prop of text, it might worth considering using children being passed down to the component so that we can use our button in a more HTML way. Here is an example of what I mean:

`export default function Button({ className, children, type, ...props }) {
	return (
		<button className={className} type={type} {...props}>
			{children}
		</button>
	);
}`

That way, we can use our new component like:

`<Button>Subscribe</Button>` instead of `<Button text="Subscribe" />`

components/TwoColumn.js
```diff
@@ -0,0 +1,25 @@
+import styles from '../styles/TwoColumn.module.scss';
+import buttonStyles from '../styles/Button.module.scss';
+import Button from './Button';
+
+export default function TwoColumn(props) {
+  const { image, title, content, rowOrder } = props;
+  return (
+    <div className={styles.container}>
+      <div className={styles.inner} style={{ flexDirection: rowOrder }}>
+        <div className={styles.itemLeft}>
```
manecipoglu, line 10: We should consider changing some of the divs here to a more semantic tag like article and section.

styles/Button.module.scss
```diff
@@ -0,0 +1,23 @@
+@import 'variables';
```
manecipoglu, line 1: @use syntax is encouraged by the Sass team instead of @import. You can find the details in their documentation.

styles/Button.module.scss
```diff
@@ -0,0 +1,23 @@
...
+
+.btn {
+  padding: 8px 32px;
+  color: #fff;
+  background-color: #292929;
+  border-radius: 32px;
+  transition: 0.3s ease;
+  cursor: pointer;
+  font-family: 'Assistant';
```
manecipoglu, line 10: Since the font-family for body is defined as Assistant, you might not need to repeat that in here.

manecipoglu:
Hi Shunya! Great first PR, congratulations.

I went too picky with my review. So absolutely do not let this discourage you. The only reason why I did that is to help you if I can. Everything is open to discussion and nobody writes perfect code. It is always encouraged to get a peer review and if only I could tell you the times I looked at my own code and thought: "What was I thinking!?". We are here to help and elevate each other. 

Let's keep working on the project.

components/Button.js
mariana-caldas: I agree with @manecipoglu .

components/Button.js
mariana-caldas: Good point, @manecipoglu 
That is especially valid for internal links. 
Please check this reference -> https://nextjs.org/docs/api-reference/next/link, @AMSdesigns, and let us know if you have any questions

components/Button.js
mariana-caldas: Could we try Muammer's approach here and you let us know how it works, @AMSdesigns , pls? Thanks!

components/TwoColumn.js
mariana-caldas: Actually, "article" would be semantic for blog posts here, @manecipoglu and @AMSdesigns , which is not the case. We could use "section" for the `.container` block wrapper tough, but this is a detail that I will leave for you to try and decide, Shunya. Thanks!

styles/Button.module.scss
mariana-caldas: Agree with Muammer.

styles/Button.module.scss
mariana-caldas: I guess it is safe to remove it here and after @manecipoglu has his PR done (with the Google fonts added), we can revisit it.

styles/Button.module.scss
mariana-caldas, line 1: Can we pls convert all the pixel values into rem for both "Button" and "Two Column" components stylesheet, @AMSdesigns ? -> https://nekocalc.com/px-to-rem-converter

styles/TwoColumn.module.scss
```diff
@@ -0,0 +1,66 @@
+@import 'variables';
```
mariana-caldas, line 1: Can we pls convert all the pixel values into rem for both "Button" and "Two Column" components stylesheet, @AMSdesigns ? -> https://nekocalc.com/px-to-rem-converter

styles/TwoColumn.module.scss
```diff
@@ -0,0 +1,66 @@
...
+  align-items: center;
+  margin: 62px 0 40px;
+}
+
+.img {
+  object-fit: cover;
+  height: 100%;
+  width: 100%;
+  max-width: 409px;
+  max-height: 545px;
```
mariana-caldas, line 53: Can we pls add a 0.25 rem "border radius" as per the mockup to this image, @AMSdesigns ? Thanks!

components/TwoColumn.js
```diff
@@ -0,0 +1,25 @@
...
+          <h2 className={styles.title}>{title}</h2>
+          <p className={styles.content}>{content}</p>
+          <Button
+            text="Learn more"
+            link="/about-us"
+            className={buttonStyles.btn}
+          />
+        </div>
+        <div className={styles.itemRight}>
+          <img src={image} alt="" className={styles.img} />
```
mariana-caldas, line 20: Can we pls try using the `<Image />` object as recommended by Next.js instead of the `img` tag here, @AMSdesigns ? Thanks!

https://nextjs.org/docs/api-reference/next/image

mariana-caldas (changes requested):
Just some details to revisit @AMSdesigns , great jb so far!
Please let me know if you have any questions to implement them. Thanks!

components/Button.js
AMSdesigns: Thank you sos much @manecipoglu and @mariana-caldas !

I am a bit unsure about how I should implement here actually.

I think we can use button as Muammer's approach like below, but if 'link' props is not passed then this may occur error.
`export default function Button(props) {
  const { className, link, type, children } = props;
  return (
    <button
      className={className}
      type={type}
      onClick={() => (location.href = `${link}`)}
    >
      {children}
    </button>
  );
}`

So should I use Link as below??
`export default function Button(props) {
  const { className, link, children } = props;
  return (
    <Link href={link}>
      <a className={className}>{children}</a>
    </Link>
  );
}`

I'd appreciate if you could tell me which is better.. thank you!

components/TwoColumn.js
AMSdesigns: Thank you, I think I just use 'section'  the .container block here if okay!

components/Button.js
manecipoglu: No, you are right to be confused. I am also debating with myself now to figure out the best approach. 

`<button>` is meant to be used for form submissions or for example, sending API requests. 
But here in your case, we want the user to navigate to another page. We should definitely use the `<Link>` component.
Here is a pretty nice article about the topic: https://css-tricks.com/use-button-element/

The problem arise here because we most likely want both our navigational links and buttons to look and feel the same way, hence we want to create a component to reuse. If it was me, I would probably change the component name to be something other than Button and make it return a Link for this issue.

components/TwoColumn.js
manecipoglu: I think this is one of the topics that has so much gray area that it is hard to be 100% sure which tag to use where. Here is my roadmap:

1) div elements should be a last resort. They  should be used for styling purposes mostly as they have no semantic meaning. 

2) articles and sections should be used to meaningfully partition a webpage. A very good indicator is heading tags like `<h1>`. If a section or article does not contain a heading, I should reconsider if I am using the right tag while keeping in mind that it is not a rule set in stone and there are exceptions. 

3) deciding between using article and section is tough. I would say article is better suited for "cards" or blog posts or comments. section is less self containing. There are no definitive rules as to which can encapsulate which. You can nest articles inside sections and sections inside articles. 

I guess, at the end of the day it comes down to the developer's taught process. @AMSdesigns, it's up to you :)
If you ask my opinion, I would completely agree with Mariana and you that section is the better choice here.

components/Button.js
mariana-caldas: That's a great point, @AMSdesigns !
I think you're right and the most straightforward way to simplify that in my opinion is to rename this component to `ButtonLink.js` since `Link` is already a Next.js object and use your second approach for now:

`export default function ButtonLink(props) { const { className, link, children } = props; return ( <Link href={link}> <a className={className}>{children}</a> </Link> ); }`

In the future, when we have real buttons that require callback props, then we create something as you suggested in your first option. Does that make sense, @AMSdesigns and @manecipoglu ?

AMSdesigns: Thank you @manecipoglu @mariana-caldas !

For this issue then, I think I will change the name of the component to `ButtonLink` as Mariana's suggestion.

components/TwoColumn.js
AMSdesigns: Thank you very much for sharing your knowledges!
I considered this by myself, and here I will go with using 'section' as above.

AMSdesigns:
well noted, thank you @mariana-caldas!

AMSdesigns:
@mariana-caldas @manecipoglu 
I submitted new commits now and it's ready for review.
Please check, thank you!

mariana-caldas (approved):
Looks good to me, @AMSdesigns ! 🥳 
All yours to merge 🚀

PR #80, opened 2022-02-24
Fix scroll-x bug and update README

mariana-caldas:

#### What is this change?

I've noticed we had a weird scroll-x caused by inadequate use of the wrapper containers and I decided to fix it. Also, the README and the PR template are being updated.

_Before_

scroll-fix

_After_

image

No.

Locally and by checking the Netlify preview.

After review.


panoramine (approved):
Great bug fix and nice updates

Issue #81, opened 2022-03-01
Adapt TwoColumn.js component to add Homepage Non Profit section


mariana-caldas:
As per [the mockup](https://www.figma.com/file/Fp0lLB5OWZLJseTs2svGbg/Web-Dev-Path_design?node-id=80%3A479), we need to create the Non Profit section. We can adapt the [TwoColumn component](https://github.com/Web-Dev-Path/web-dev-path/blob/master/components/TwoColumn.js) to use it for that section as well. To do so, you need to:

* allow the component to receive props for the `background-color` and text/content `color`
* use [conditional rendering](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator) for the `inner__image` section in a way that the section is only displayed when there is an image added to the component.

This issue is related to #77 , but it's not a problem to finish it without the respective decoration because it can be positioned later.

Once you have the most up-to-date master branch locally (`git pull` from your local master branch), create a new branch `git checkout -b feature/add-nonprofit-homepage` to get started.

![non-profit-section](https://user-images.githubusercontent.com/20016897/156091478-f8046958-36aa-42a7-8f7e-45e68c8ee73b.JPG)

**Acceptance criteria:**

* Add the new section by adapting the `TwoColumn.js` components with the new required logic. Make sure the other section still works as intended.

* Make sure the component is responsive on mobile by stacking when the screen is < 768px

* Update the `CHANGELOG.md` file

PR #82, opened 2022-03-01
Updated TwoColumn.js (Nonprofit homepage)

vmcodes:
| [**81**](https://github.com/Web-Dev-Path/web-dev-path/issues/81) |

Yes

#### What is this change?
Updated the TwoColumn.js component to receive props for font color, background color, button color, and button background color. Conditional display of any images passed through as a prop was also added.

Had to update ButtonLink.js to accept style props as well, this was to make sure the default button colors defined in the .scss files could be overridden if needed. 

Since the initial PR, I've added an altTag and linkText prop to the TwoColumn component, as well as the requested scss exports required to pass in styles to the component.

Tested locally and made sure that any added props worked as intended without errors, as well as if a color, linkText, or image prop was not passed through.

This should be ready to be merged after review.

### DIFF
--- a/components/TwoColumn.js
+++ b/components/TwoColumn.js
@@ -5,26 +5,47 @@ import buttonStyles from '../styles/ButtonLink.module.scss';
 
 export default function TwoColumn(props) {
   // Add rowOrder="row-reverse" prop to the component to reverse its order on desktop
-  const { image, title, content, rowOrder } = props;
+  const { image, title, content, rowOrder, color, bgColor, btnColor, btnBg } =
+    props;
+
+  const styleProps = {
+    btn: {
+      color: btnColor,
+      backgroundColor: btnBg,
+    },
+    container: {
+      color: color,
+      backgroundColor: bgColor,
+    },
+  };
+
   return (
-    <section className={styles.container}>
+    <section className={styles.container} style={styleProps.container}>
       <div className={styles.inner} style={{ flexDirection: rowOrder }}>
         <div className={styles.inner__content}>
-          <h2 className={styles.title}>{title}</h2>
+          <h2 className={styles.title} style={{ color: color }}>
+            {title}
+          </h2>
           <p className={styles.content}>{content}</p>
-          <ButtonLink link="/about-us" className={buttonStyles.btn}>
+          <ButtonLink
+            link="/about-us"
+            className={buttonStyles.btn}
+            styles={styleProps.btn}
+          >
             Learn more
           </ButtonLink>
         </div>
-        <div className={styles.inner__image}>
-          <Image
-            src={image}
-            alt=""
-            className={styles.img}
-            layout="fill"
-            objectFit="cover"
-          />
-        </div>
+        {image && (
+          <div className={styles.inner__image}>
+            <Image
+              src={image}
+              alt=""
+              className={styles.img}
+              layout="fill"
+              objectFit="cover"
+            />
+          </div>
+        )}
       </div>
     </section>
   );
--- a/components/TwoColumn.js
+++ b/components/TwoColumn.js
@@ -5,8 +5,19 @@ import buttonStyles from '../styles/ButtonLink.module.scss';
 
 export default function TwoColumn(props) {
   // Add rowOrder="row-reverse" prop to the component to reverse its order on desktop
-  const { image, title, content, rowOrder, color, bgColor, btnColor, btnBg } =
-    props;
+  const {
+    image,
+    altTag,
+    title,
+    content,
+    rowOrder,
+    color,
+    bgColor,
+    btnColor,
+    btnBg,
+    link,
+    linkText = 'Learn more',
+  } = props;
 
   const styleProps = {
     btn: {
@@ -28,18 +39,18 @@ export default function TwoColumn(props) {
           </h2>
           <p className={styles.content}>{content}</p>
           <ButtonLink
-            link="/about-us"
+            link={link}
             className={buttonStyles.btn}
             styles={styleProps.btn}
           >
-            Learn more
+            {linkText}
           </ButtonLink>
         </div>
         {image && (
           <div className={styles.inner__image}>
             <Image
               src={image}
-              alt=""
+              alt={altTag}
               className={styles.img}
               layout="fill"
               objectFit="cover"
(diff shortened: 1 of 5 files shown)


pages/index.js
```diff
@@ -20,6 +20,15 @@ export default function Home() {
         content="The Web Dev Path is a team of professional developers project that aims to provide a comprehensive path for people who seek to begin their web development journey. "
       />
 
+      <TwoColumn
+        title="Nonprofit?"
+        content="Web Dev Path can help your nonprofit with web projects of various sizes. Connect with us to find out how."
+        color="#ffffff"
+        bgColor="#292929"
+        btnColor="#292929"
```
manecipoglu, line 28: I believe we have this color defined in our variables partial but because we are passing them to the component as props, I don't think we can pull them easily like we do with the scss files. Is there a way to do that? 

I realize the issue is asking for changeable props for background color and text color but maybe a solution can be found by passing a className as prop instead of passing the styles. This way we can apply the "separation of concerns" principles. This has the downside of compiling more css than necessary whenever we want to use TwoColumns component with different styling. What do you think @mariana-caldas ?

components/TwoColumn.js
```diff
@@ -5,26 +5,47 @@ import buttonStyles from '../styles/ButtonLink.module.scss';
...
-        <div className={styles.inner__image}>
-          <Image
-            src={image}
-            alt=""
-            className={styles.img}
-            layout="fill"
-            objectFit="cover"
-          />
-        </div>
+        {image && (
```
manecipoglu, line 38: Good use of conditional rendering here.

manecipoglu (commented):
Very nice PR Vincent. Thank you.

components/TwoColumn.js
```diff
@@ -5,26 +5,47 @@ import buttonStyles from '../styles/ButtonLink.module.scss';
 
 export default function TwoColumn(props) {
   // Add rowOrder="row-reverse" prop to the component to reverse its order on desktop
-  const { image, title, content, rowOrder } = props;
+  const { image, title, content, rowOrder, color, bgColor, btnColor, btnBg } =
+    props;
+
+  const styleProps = {
```
mariana-caldas, line 11: It's an elegant solution to add all those style props in a single object, @vmcodes ! Well done!

components/TwoColumn.js
```diff
@@ -5,26 +5,47 @@ import buttonStyles from '../styles/ButtonLink.module.scss';
...
       <div className={styles.inner} style={{ flexDirection: rowOrder }}>
         <div className={styles.inner__content}>
-          <h2 className={styles.title}>{title}</h2>
+          <h2 className={styles.title} style={{ color: color }}>
+            {title}
+          </h2>
           <p className={styles.content}>{content}</p>
-          <ButtonLink link="/about-us" className={buttonStyles.btn}>
+          <ButtonLink
+            link="/about-us"
```
mariana-caldas, line 31: Could you please create a prop to pass the `<ButtonLink />` string link dynamically when calling the component as well, @vmcodes ? It's possible that we're going to use the `TwoColumns.js` with a different link in other situations. Thanks!

pages/index.js
mariana-caldas: This is one of the unique cases the use of inline styling makes sense to me, @manecipoglu. Because we're just changing colors and the idea is to reuse this component as much as we need, I think it is fine to keep specific props as is. However, great point on highlighting the use of variables here, Muammer! I agree with you.

Could you please follow this [step 3 tutorial ](https://spacejelly.dev/posts/how-to-export-sass-variables-to-javascript-with-next-js/)to export the color variables from our [variables file](https://github.com/Web-Dev-Path/web-dev-path/blob/master/styles/_variables.scss) and import them here, @vmcodes ? Please let m know if you need any assistance with that! Thanks :)

mariana-caldas (changes requested):
This is such a great PR, @vmcodes ! I only see small adjustments. Please let m know if you need any help to implement them and after you commit those changes, could you please [request our review again](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review) ? Thank you! 🚀

pages/index.js
manecipoglu: > Could you please follow this [step 3 tutorial ](https://spacejelly.dev/posts/how-to-export-sass-variables-to-javascript-with-next-js/)to export the color variables from our [variables file](https://github.com/Web-Dev-Path/web-dev-path/blob/master/styles/_variables.scss) and import them here

This is exactly what I wanted to do instead. I wasn't aware of the :exports syntax. Good to know. Thanks for the link!

styles/TwoColumn.module.scss
```diff
@@ -63,3 +63,9 @@
     }
   }
 }
+
```
mariana-caldas, line 66: The idea of exporting the colors is correct, @vmcodes , however, we should do that from our `variables.scss` file  instead -> https://github.com/Web-Dev-Path/web-dev-path/blob/master/styles/_variables.scss

mariana-caldas (changes requested):
We only need to export the color variables from the right file (https://github.com/Web-Dev-Path/web-dev-path/blob/master/styles/_variables.scss) to finish this PR, @vmcodes!
We're getting close! 🚀

styles/TwoColumn.module.scss
vmcodes: I received the following error when trying to do that, which is why I moved them to TwoColumn.module.scss.

./styles/_variables.scss
Global CSS cannot be imported from files other than your Custom <App>. Due to the Global nature of stylesheets, and to avoid conflicts, Please move all first-party global CSS imports to pages/_app.js. Or convert the import to Component-Level CSS (CSS Modules).
Read more: https://nextjs.org/docs/messages/css-global
Location: pages\index.js

vmcodes: I should have listed that under complications.

manecipoglu: I am not sure but can you try without the underscore "_" ? Like this: ./styles/variables.scss

mariana-caldas: Oh, I see. That makes sense! Thanks for checking that, @vmcodes !

mariana-caldas (approved):
LGTM, @vmcodes ! Great job!
It's all yours to [merge](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/merging-a-pull-request) :whenever you have a moment!

Issue #77, opened 2022-02-11
Add decoration components (YellowBracket and BlueStick)


mariana-caldas:
As per [the mockup](https://www.figma.com/file/Fp0lLB5OWZLJseTs2svGbg/Web-Dev-Path_design?node-id=80%3A479), we need to create two components from the attached SVG files that will be initially used on the Homepage, but, as expected of any component, should be built in a way we can reuse it in any part of the platform in the future.

Reference: https://www.sanity.io/guides/import-svg-files-in-react#c125367dce7c

Once you have the most up-to-date master branch locally (`git pull` from your local master branch), create a new branch `git checkout -b feature/add-decorations` to get started.

![blue-stick](https://user-images.githubusercontent.com/20016897/153531271-8921df61-050b-4436-bd7b-1dce14ec8b44.svg)

![yellow-bracket](https://user-images.githubusercontent.com/20016897/153531275-2114b7cb-d83c-48c8-a5e7-dc28e0fc4ba6.svg)

**Acceptance criteria:**

* Create two components called `YellowBracket.js` and `BlueStick.js` that will receive a `className` `prop` (for styling the position according to the section) and test how they scale on different screens by adding them to the homepage. You don't need to think about any proper styling for now since it will be defined according to the class added to it depending on the section.

* Update the `CHANGELOG.md` file

PR #84, opened 2022-03-03
Feature/add decorations

vmcodes:
| [77](https://github.com/Web-Dev-Path/web-dev-path/issues/77) |

Yes.

#### What is this change?

Adding decoration components YellowBracket.js and BlueStick.js. 
Updated to Bracket.js and Stick.js due to default props added for fill. Also, I added the components to the homepage along with the missing TwoColumn.js component to allow for a preview of the positioning of these components.

Please see the Netlify preview when possible.

The svg colors are unable to be changed directly via scss, so the Image component provided by Next JS appears to not be the best option in this case. I added an optional 'fill' prop that has a default value of the original fill color. However, the className prop works as expected when a class is passed through as a prop.

Tested locally with different scss values and props. The lack of a prop had no negative effects, but did show that the svg could be modified. Different scss values passed through the className prop also worked as expected for positioning and changing height/width.

Before committing I pulled from master, so this branch should be ready to be merged after review.


mariana-caldas:
I'm converting this PR as Draft because it seems it wasn't created from the most updated master, the version that has the "Nonprofit section". As soon as you have it solved, @vmcodes , could you please click over "Ready for review", assign yourself for this PR, and request reviews?

vmcodes:
> 

I did pull from master first before pushing the recent commits.

mariana-caldas:
> @mariana-caldas I did pull from master first before pushing the recent commits, please see the Netlify deploy preview.

Your CHANGELOG file said otherwise, @vmcodes  -> it did not include the previous feature information and I found it intriguing. 
Also, there are more details we talked about on Slack that make this PR incomplete for review. Please only request reviews when your PR is completely ready, does that make sense? Please let me know if you need help to implement the other remaining changes. Thanks!

mariana-caldas (approved):
This PR is great, @vmcodes !
Your careful approach with the issue instructions and the fact you leverage the use of the TwoColumn.js component to position the new decoration components is impressive! Great job! 🚀

PR #86, opened 2022-03-14
Update navbar with scrolly effect

manecipoglu:
| [**83**](https://github.com/MarianaSouza/web-dev-path/issues/83) |

Yes.

#### What is this change?
This adds an intersection api to our navbar. With the current configuration, the navbar re-appears after the header is no longer in viewport view and more than 300px is scrolled after that. There are also changes made to the styling for mobile and desktop views. Previous implementation of hamburger menu is also replaced.  

Too many to count :) 
But after experimenting thoroughly and thoughtfully, I reversed a lot of the make-shift solutions I made.

A simple one was using the navLinks we have inside the utils.js. The order and wording does not match some of the links we use at our footer. I believe we can easily change our style guide for footer to better accommodate both situations and the order could be changed easily from css since we use flexbox, not to mentioned maybe making two arrays for our links is another solution. However, since we also want to use our ButtonLink component, rather than having an if\else condition inside the map function we put inside our return, writing the nav items in our JSX seemed like the more clear and easier to read solution. We do lose our ability to change the links from only one location in the future with this though. Any ideas?

Another simple one was implementing the intersection observer. Since the biggest selling point of using this api is to reduce the load on the browser, managing this intersection with the help of the state seemed counter-intuitive since changing the state would trigger a re-render of the component. Hence, I decided to go with refs. 

My main discomfort with this change is the CSS code I came up with. Especially with the sticky class implementation, it feels like this is fast turning into a spaghetti code. 

Since we do not have a very distinct header, I used an arbitrary height based on logo size but I believe shortening the height would make it look better, especially once it gets sticky.

Locally.

After code review.


vmcodes:
@manecipoglu I think that the mobile views look great, but maybe the height of the desktop view could be a bit smaller. @mariana-caldas I do think this may be a particularly good use case for the [clsx](https://www.npmjs.com/package/clsx) library. While I realize it hasn't been updated in some time, it has become somewhat of a go to considering the download count for situations like the navbar color change.

```<ul className={`${styles.nav__links} ${active ? styles.active : ''}`}>```

mariana-caldas:
Hey @manecipoglu ! Great job with the use of the useRef() hook here. I agree that is one of the unique cases it makes sense to use it instead of the regular state since we don't need to keep changing the DOM, right? 🚀 🚀 

Now, about the questions:

1) As we already talked about in your last PR with the footer, when you're facing a lot of questions like that, you don't have to make all the decisions by yourself. In this case, as the previous one, you should have reached out to your team lead and to a designer, such as @rebecatschiedel , to discuss some possibilities for the header height, for example. You'll see by reading the comments below that would be much easier for you to understand everything and even for myself if we had just talked about that on Slack whenever you felt stuck.

2) Since we're having to add a special class to the ButtonLink for styling anyway, I think it's better to keep the map() functionality and add the last nav__item outside the loop. Doe that make sense? If not, please reach out.

3) The way you set the styling makes sense to me and since everything is isolated (every component has its own styling) I don't see us turning it into a spaghetti code. Those CSS effects usually require indeed some effort and more classes when necessary. However, the navbar needs some improvements, not so much in regards to height on the desktop since the component will have a transparent background that will be mixed with the header itself, but when adding the sticky class on scroll.

- Please remove the huge amount of padding you added to create the "container effect" and simply wrap the navbar into a container with the same styling we see on the TwoColumn.js section. I guess we could even create a component Container.js to make sure we're using the same in any section, right?

 - On scroll, when the sticky class is present, the navbar height should look like as you set it on mobile, which means not so thick. That will probably require we adapt the other links too. Try to communicate and reach out to @rebecatschiedel to pick her brain a little bit on that, OK? This is a teamwork process! 
 
 - Because some sections will probably be white, we should add some box-shadow to the sticky navbar as well. Something like we see in this example -> https://www.contentful.com/blog/. Once more, reach out to @rebecatschiedel before advancing.
 
 - Finally, I see this error when loading the project locally. Do you also see that? If so, could you please take a look at it?
 
 
image

The hardest part is already done, so just some details to sort out. Reach out if you have questions, thanks!

mariana-caldas:
> @manecipoglu I think that the mobile views look great, but maybe the height of the desktop view could be a bit smaller. @mariana-caldas I do think this may be a particularly good use case for the [clsx](https://www.npmjs.com/package/clsx) library. While I realize it hasn't been updated in some time, it has become somewhat of a go to considering the download count for situations like the navbar color change. What do you think?
> 
> `` <ul className={`${styles.nav__links} ${active ? styles.active : ''}`}> ``

I prefer we try not to use more dependencies to solve small problems, @vmcodes 🙌🏼

manecipoglu:
I implemented the changes mentioned. After discussing with @rebecatschiedel, I changed the order and the wording for the navlinks. I brought back the mapping of our navlinks on the header, added a Container component (had to forward the ref to make it work properly. I don't know if there are better ways). I also added a shadow effect for the future when we make the header background transparent and reduced the height of it when it is sticky. I am not very proud of the way I implemented CSS. Thanks for the feedback!

mariana-caldas (approved):

vmcodes (commented):
Wow @manecipoglu this looks great, particularly the mobile and tablet views. I hate to be that guy, but my only suggestion is to make the initial desktop header the same dimensions as the header after scrolling down. Also, nice work with the useRef() hook.

mariana-caldas:
> Wow @manecipoglu this looks great, particularly the mobile and tablet views. I hate to be that guy, but my only suggestion is to make the initial desktop header the same dimensions as the header after scrolling down. Also, nice work with the useRef() hook.

 I agree, @vmcodes, and the height/dimensions issue will be solved in the next feature, the Header component that will wrap this Navbar and will allow to make its background transparent with a nice image :)

Issue #78, opened 2022-02-11
Add ThreeColumn component


mariana-caldas:
As per [the mockup](https://www.figma.com/file/Fp0lLB5OWZLJseTs2svGbg/Web-Dev-Path_design?node-id=80%3A479), we need to create the ThreeColumn component that will be used initially on the Homepage, but, as expected of any component, should be built in a way we can reuse it in any part of the platform in the future.

Once you have the most up-to-date master branch locally (`git pull` from your local master branch), create a new branch `git checkout -b feature/add-three-column` to get started.

![image](https://user-images.githubusercontent.com/20016897/153533412-417871d1-37f0-460e-a40a-d004a1395be5.png)

**Acceptance criteria:**

* Create a component called `ThreeColumn.js` that will receive some properties (`props`) for the _imageOne, imageTwo, imageThree, imageTitleOne, imageTitleTwo, imageTitleThree, imageContentOne, imageContentTwo, imageContentThree_ and the _className_ for the section styling.

* Use the `<Link />` [object](https://nextjs.org/docs/api-reference/next/link) to link the image content info to `about-us` and `contact-us` pages for now.

* Make sure the component is responsive on mobile by stacking when the screen is < 768px

* Update the `CHANGELOG.md` file

**Please, keep in mind that once you assign this task to yourself you'll need to complete it in 5 days**

PR #88, opened 2022-03-18
Add three column component

vmcodes:
| [**78**](https://github.com/Web-Dev-Path/web-dev-path/issues/78) |

yes

#### What is this change?
Added ThreeColumn.js and Card.js components. SCSS files for each component were also made, as well as adding ThreeColumn.js to the homepage.

After discussing we decided to make a reusable card component instead of placing three cards within the ThreeColumn.js component.

Locally.

Should be ready after review.


components/ThreeColumn.js
```diff
@@ -0,0 +1,26 @@
+import Card from './Card';
+import styles from '../styles/ThreeColumn.module.scss';
+
+export default function ThreeColumn(props) {
+  const { images, altTags, titles, content, links, linkText } = props;
+
+  return (
+    <section className={styles.container}>
+      <div className={styles.inner__content}>
+        {titles.map((title, index) => {
```
mariana-caldas, line 10: I love your solution with a map() here, @vmcodes. Super DRY and creative!

mariana-caldas (approved):
Such a great solution, @vmcodes ! Thank you for that! 🚀 🚀 
I took over your branch just to adjust some styling that was not only related to the new components.
Could you please wait until Wednesday to get it merged so we give others a chance to review it too?

PR #91, opened 2022-03-31
Install Yarn and update script commands

mariana-caldas:

Yes.

#### What is this change?

Because [Yarn](https://yarnpkg.com/) is a _package manager_ more secure and offers better performance than NPM, we're adding that to the project. A new set of commands are being added to the `package.json` scripts as well so when we run de project locally, it opens automatically on `http://localhost:3010`

To install Yarn I took the following steps:
* I've removed the old `node_modules` folder by running `rm -rf node_modules/`
* Run `yarn install` to create the `yarn.lock` file, the 'guy' that replaced the NPM `package-lock.json` and contains all the dependencies we need.

_**Next time a dev starts a feature or chore, they will have to perform the same steps described above after pulling from `master`. Keep in mind that we're using `yarn` commands as described on the README updates from now on**_

No.

Locally and also looked at the Netlify preview link to check for console errors.

After review.


vmcodes (approved):
I think using Yarn is a great choice!

PR #96, opened 2022-04-09
Update local commands for macOS

mariana-caldas:

Yes.

#### What is this change?

The project `yarn` commands to run it locally are being updated to allow opening the project automatically on macOS and Windows. The way they were set before, only Windows users would be able to run the project locally.

It was tricky to figure out that the `start` command only worked for Windows while `open` is the proper one for macOS.

Locally by testing the project installation both on a PC and on a Mac.

After code review.


panoramine (approved):
Great PR! Now the project is more accessible.

Issue #95, opened 2022-04-08
Add the `Header.js` component to wrap the navbar


mariana-caldas:
As per the design files, [homepage](https://www.figma.com/file/Fp0lLB5OWZLJseTs2svGbg/Web-Dev-Path_design?node-id=236%3A563) and [about-us](https://www.figma.com/file/Fp0lLB5OWZLJseTs2svGbg/Web-Dev-Path_design?node-id=566%3A2257), we need to create a `Header.js` component to wrap the navbar.

Since the content and image are going to be different on the different pages, we'll need to render [the content conditionally](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator) per page by using the `useRouter` hook `pathname`. We have an example of that hook on the `Nav.js` component.

![homepage-header](https://user-images.githubusercontent.com/20016897/162370237-da0fb66e-ddb6-4883-9289-eba2c1efbbbb.png)

![about-us-header](https://user-images.githubusercontent.com/20016897/162370261-add0d30f-2a5f-4aee-8ccf-0f04bc1d4097.png)

Once you have the most up-to-date master branch locally (git pull from your local master branch), create a new branch `git checkout -b feature/add-header-component` to get started.
Check out our [Wikipedia page](https://github.com/Web-Dev-Path/web-dev-path/wiki/Starting-the-process) to learn more about the process.

**_Please, keep in mind that once you assign this task to yourself you'll need to complete it in 15 days._**

---

**Approach suggestion:**

* The component should have `props` that allow to set the `title` , `content`, and a `customClassname` that can be used to set the title color depending on the the page. The homepage also requires we add an special `prop`, the `title_dynamic`, that will going to be a yellow `<span></span>` with content that will change dynamically in the future.

* The component will wrap the `Nav.js` on `Layout.js`

* The component should have a 100% height as per[ this reference site](https://mariana-caldas.github.io/bankist/)

* Create a `JSON` or `js` file under `utils`  where the pages' content, the `props` content, will be added. Map over that file on `Layout.js`, where the `Header.js` will be referred and render [the content conditionally](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator) per page by using the `useRouter` hook `pathname`. We have an example of that hook on the `Nav.js` component.

---

**Acceptance criteria:**

* Once we access the [homepage](https://www.webdevpath.co/), we need to see the correct header as per mockup, which means its proper image and content without the button.

[...]

PR #98, opened 2022-04-18
Add hero component

mariana-caldas:
| [**95**](https://github.com/Web-Dev-Path/web-dev-path/issues/95) |

Yes.

#### What is this change?

The `Hero.js` component, the one that wraps the `Nav.js`, is being added to the project. The component allows for setting image background and content per page by considering its pathname. 

Adding the dynamic content on the homepage effect was tricky. To do so, I used a solution that combined `useState()` to set the `dynamicTitles` array index and `useEffect()` where the titles are updated every 1350 milisseconds.

Locally and by checking the Netlify preview link.

After code review.

### DIFF
--- a/components/Layout.js
+++ b/components/Layout.js
@@ -1,4 +1,5 @@
 import { useRouter } from 'next/router';
+import { heroOptions } from '../utils/hero-options';
 import Hero from './Hero';
 import Meta from './Meta';
 import Footer from './Footer';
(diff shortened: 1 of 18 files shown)


panoramine (approved):
Wow! Learned a lot reviewing theses changes.

components/Layout.js
```diff
@@ -1,15 +1,25 @@
...
+import { useRouter } from 'next/router';
+import { heroOptions } from '../utils/hero-options';
+import Hero from './Hero';
 import Meta from './Meta';
 import Footer from './Footer';
 import styles from '../styles/Layout.module.scss';
 
 export default function Layout({ children }) {
+  const router = useRouter();
+  const heroPathKeys = Object.keys(heroOptions)
```
cherylli, line 10: is this block of code (sort and filter) for something else? 
`<Hero {...heroOptions[router.pathname]} />` works 

```suggestion
  const router = useRouter();
  return (
    <>
      <Meta />
      <main className={styles.main}>
        <Hero {...heroOptions[router.pathname]} />
        {children}
        <Footer />
      </main>
    </>
  );
```

cherylli (commented):
Just one block of code I'm unsure of
have tested locally

components/Layout.js
mariana-caldas: Yes, that part is needed there, @cherylli .
It is a rule created to make sure all the future blog pages are initially directed to the primary blog hero we see on line 26 of `hero-options.js`
On line 10, I'm grabbing the object keys so that the object elements are always sorted out and filtered by their initial part ( I recommend consoling log the "k" on line 12). 
Then, on line 14, you see the heroKey part to be considered will always be **the last** part of that object key in case I want to overwrite the condition by adding a new "blog" property with a more specific path, such as `blog/my-post-title`.

Issue #100, opened 2022-05-03
Reveal sections on scrolling with Intersection Observer API


mariana-caldas:
As a user, I want to see all the sections but the `Hero.js` being revealed only on scrolling and with a `translateY` effect that simulates a smooth movement towards the top as per [this reference website](https://mariana-caldas.github.io/bankist/). [This](https://github.com/mariana-caldas/bankist) is the reference website codebase.

Once you have the most up-to-date `master` branch locally (git pull from your local master branch), create a new branch git checkout -b feature/reveal-sections-scrolling to get started. Check out our [Wikipedia page](https://github.com/Web-Dev-Path/web-dev-path/wiki/Starting-the-process) to learn more about the process.

**Please, keep in mind that once you assign this task to yourself you'll need to complete it in 10 days.**

**Approaches' suggestions:**

1. [Intersection Observer API with useRef, useState, and UseEffect](https://spacejelly.dev/posts/how-to-trigger-a-function-when-scrolling-to-an-element-in-react-intersection-observer/)
2. [Use a library, React Intersection Observer](https://www.npmjs.com/package/react-intersection-observer)

**Acceptance Criteria:**

- Make sure all the homepage sections but the hero are displayed only on scrolling as per [the reference website](https://mariana-caldas.github.io/bankist/) with no console errors both on desktop and mobile. 
- Test the feature in many screen sizes, you can use the Inspect tool for that.
- Test the feature in many browsers, such as Chrome, Firefox, Edge, and Safari (MAC).
- Update the CHANGELOG.md file.

PR #101, opened 2022-05-13
Feature/reveal sections scrolling

cherylli:
| [**100**](https://github.com/Web-Dev-Path/web-dev-path/issues/100) |

yes

#### What is this change?
 all the sections but the Hero.js being revealed only on scrolling and with a translateY effect that simulates a smooth movement towards the top

no

run locally on firefox, chrome, brave and in responsive design mode

after reviews

### DIFF
--- a/components/Footer.js
+++ b/components/Footer.js
@@ -4,42 +4,39 @@ import { linksNav } from '../utils/links';
 import NewsletterSubscribe from './mailchimp/NewsletterSubscribe';
 import Image from 'next/image';
 import Container from './Container';
-import RevealContentContainer from './RevealContentContainer';
 
 export default function Footer() {
   return (
-    <RevealContentContainer>
-      <footer className={footerStyles.footer}>
-        <NewsletterSubscribe />
-        <Container className={footerStyles.footer__inner}>
-          <nav className={footerStyles.footer__nav} aria-label="Main">
-            <ul className={footerStyles.footer__navList}>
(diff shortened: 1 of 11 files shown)


cherylli:
I have added interestion observer (with a custom hook) on TwoColumns and CardColumns. Which means everything with TwoColumns and CardColumns will have this revealing effect, which works fine for now. I'm not sure if this will affect future contents if these components are to be used elsewhere. 

Another way I can think of is create another component with intersect to wrap around individual front page components

components/Footer.js
```diff
@@ -4,39 +4,42 @@ import { linksNav } from '../utils/links';
 import NewsletterSubscribe from './mailchimp/NewsletterSubscribe';
 import Image from 'next/image';
 import Container from './Container';
+import RevealContentContainer from './RevealContentContainer';
```
mariana-caldas, line 7: Could you please remove the effect from the footer? Thanks!

mariana-caldas (changes requested):
I just have that single request on the footer, @cherylli !
This looks great, thanks!

components/Footer.js
cherylli: removed

mariana-caldas (approved):
LG(reat)TM, @cherylli ! 🥳 
Can we just wait a few days to see if someone else has more feedback before you merge it, pls? Thanks!

Issue #102, opened 2022-05-26
Create Slider.js component for CardsColumns.js on mobile


mariana-caldas:
As a user, I want to see the cards from the `CardsComponent.js` displayed in a slider on mobile and tablets (screens <  769px) as per [this reference website](https://thecozyshop.com/) (and attached)

![carousel-example-mobile](https://user-images.githubusercontent.com/20016897/170416708-7c63e2af-ed78-4815-b4c2-0d56c7580d4a.png)

![carousel-example-tablet](https://user-images.githubusercontent.com/20016897/170416711-d1160f5e-eabd-4b93-8d8b-9d2d164b8566.png)

---

Once you have the most up-to-date `master` branch locally (git pull from your local master branch), create a new branch `git checkout -b feature/implement-slider` to get started. Check out our [Wikipedia page](https://github.com/Web-Dev-Path/web-dev-path/wiki/Getting-Started) to learn more about that process.

**Please, keep in mind that once you assign this task to yourself you'll need to complete it in 10 days.**

**_Approach suggestion:_**

[Use the component from this Flickity React library](https://www.npmjs.com/package/react-flickity-component)

**_Acceptance Criteria:_**

- Ensure the CardsColumns.js is displayed in a slider on tablets and mobile as per [the reference website](https://thecozyshop.com/) on the Homepage with no console errors.
- Test the feature in many screen sizes, you can use the Inspect tool for that.
- Test the feature in many browsers, such as Chrome, Firefox, Edge, and Safari (MAC).
- Update the CHANGELOG.md file.

luke-buttifant:
Hey, I would like to implement this!

PR #103, opened 2022-06-02
added card carousel (Home page)

luke-buttifant:
| [**102**](https://github.com/Web-Dev-Path/web-dev-path/issues/102) |

Changelog has been updated. 

#### What is this change?
Added a card carousel to the home page. 

No!

I have tested various different screen sizes and there is also no console errors. 

After review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Header.js component
 - reCAPTCHA
 - useIntersect Hook, Reveal container for scrolling using intersection API
+- A new swipable feature on CardColumns.js
 
 ### Fixed
 
@@ -38,6 +39,4 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - normalized buttons' and links' styling
 - updated `package.json` commands for macOs
 
-### changed
 
-- Prompt cards are now swipable. (Home page)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,3 +37,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - installed yarn
 - normalized buttons' and links' styling
 - updated `package.json` commands for macOs
+
+### changed
+
+- Prompt cards are now swipable. (Home page)
--- a/components/CardsColumns.js
+++ b/components/CardsColumns.js
@@ -1,7 +1,8 @@
 import { Swiper, SwiperSlide } from 'swiper/react';
-import { Pagination } from 'swiper';
+import { Pagination, Navigation } from 'swiper';
 import 'swiper/css';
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -37,3 +37,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - installed yarn
 - normalized buttons' and links' styling
 - updated `package.json` commands for macOs
+
+### changed
+
+- Prompt cards are now swipable. (Home page)
```
cherylli, line 43: maybe better to mention cards under which section or component. Should it be under #Added? a new swipable feature?

components/CardsColumns.js
```diff
@@ -1,6 +1,8 @@
 import Card from './Card';
 import Container from './Container';
+import { Swiper, SwiperSlide } from 'swiper/react';
```
cherylli, line 3: The import order convention is usually node_module public module -> private modules -> local files

components/CardsColumns.js
```diff
@@ -11,22 +13,46 @@ export default function CardsColumns({
...
-              content={content[index]}
-              link={links[index]}
-              linkText={linkText[index]}
-              key={index}
-            />
-          );
-        })}
-      </div>
-    </Container>
+    <>
```
cherylli, line 16: this <></> may not be necessary.

cherylli (commented):
Works perfectly! Just a few comments

General comments about the PR 
Please fill in "IssueNumber"
And assign yourself to the PR, and probably put tags in

There's no indication of more cards on the right side when viewing, so users probably won't notice it's a slider and just keep scrolling down.
[](https://i.imgur.com/cUqdJVK.png)

CHANGELOG.md
luke-buttifant: Ok sure that makes more sense!

components/CardsColumns.js
luke-buttifant: Ok no problem 👍🏼

cherylli:
Hi, I've just noticed the slider is still active at up to 1322 px width, but it should not be showing on screens > 769px,  as per requirement. 

I would decrease the gap between the dots and cards a little as they are off screen on my mobile.

luke-buttifant:
> Hi, I've just noticed the slider is still active at up to 1322 px width, but it should not be showing on screens > 769px, as per requirement.
> 
> I would decrease the gap between the dots and cards a little as they are off screen on my mobile.

To have 3 cards showing at 769px caused them to bunch up, like this: 

image

So I played with the breakpoints and how many cards were showing to make it more seamless. The other alternative would be to make the cards/margin smaller so that they can all fit at 769px. 

I can change the dots, or I could remove them completely and instead have a slight bit of the next card showing to prompt a swipe instead. Like shown: 
image

What do you think?

cherylli:
oh yeah that does look terrible. I think the dots are fine, can they be moved up or is the style fixed?

PR #112, opened 2022-07-04
Fix ButtonLink.js styles

mariana-caldas:

Yes.

#### What is this change?

The `ButtonLink.js` styles are being adjusted to remove the underline and opacity when the link added to the `TwoColumn.js` is not a button. We can see the problem on the homepage live website:

image

Kinda, I almost fell on the `!important` hell 🤣 , but I had a brain to pick thank Gaia! Thanks, @cherylli 👩🏼‍🎤 

Locally and by using the Netlify preview link.

After code review.

### DIFF
--- a/styles/ButtonLink.module.scss
+++ b/styles/ButtonLink.module.scss
@@ -13,10 +13,11 @@
   min-width: 16rem;
   display: inline-block;
   border: 1px solid $primary-content-color;
+  text-decoration: none !important;
   @include transition(all 0.3s ease);
 
   &:hover {
-    text-decoration: none;
+    opacity: 1 !important;
   }
 
   &.inverted-grey {
--- a/styles/ButtonLink.module.scss
+++ b/styles/ButtonLink.module.scss
@@ -13,11 +13,10 @@
   min-width: 16rem;
   display: inline-block;
   border: 1px solid $primary-content-color;
-  text-decoration: none !important;
   @include transition(all 0.3s ease);
 
   &:hover {
-    opacity: 1 !important;
+    text-decoration: none;
   }
 
   &.inverted-grey {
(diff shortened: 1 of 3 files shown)


styles/ButtonLink.module.scss
```diff
@@ -13,10 +13,11 @@
   min-width: 16rem;
   display: inline-block;
   border: 1px solid $primary-content-color;
+  text-decoration: none !important;
```
cherylli, line 16: A way to avoid using !important here is put `text-decoration: underline` under about-content, so it's specific to the TwoColumn component in the about page, like so,

```
&.about-content {
     {stuff}
      .inner__content {
        a {
          text-decoration: underline;
        }
```

[](https://i.imgur.com/FhCPZjG.png)

styles/ButtonLink.module.scss
```diff
@@ -13,10 +13,11 @@
...
   @include transition(all 0.3s ease);
 
   &:hover {
-    text-decoration: none;
+    opacity: 1 !important;
```
cherylli, line 20: this can probably be dealt with in a similar way via the customInnerClass of TwoColumn.js

cherylli (commented):
my suggestions. not sure which one you prefer.

Faith-Nchifor (approved):
This is good!

styles/ButtonLink.module.scss
mariana-caldas: That makes sense! I'll try that. Thanks, @cherylli , you're the best! 🚀

mariana-caldas:
Your suggestion gave me an even better idea, @cherylli ! Thanks for that! 🚀

cherylli (approved):
looks good!

PR #130, opened 2022-08-30
Adding "Wanna learn more?" and "How to get started?" sections to the About Us page

cherylli:
| [**125**](https://github.com/MarianaSouza/web-dev-path/issues/125) |

yes

#### What is this change?
1. Add wanna learn more part in about page
2. modify TwoColumn to accept a second column of text (or other components) instead of an image

no

run locally in desktop and mobile view

after review


pages/about.js
```diff
@@ -127,6 +127,50 @@ export default function AboutUs() {
...
+            }
+              rowOrder='row'
+              linkText='Contact us'
+              link='/contact'
+              customBtnClass='inverted-grey'
+              color={primary}
+              bgColor={lightBgColor}
+              customInnerClass='second-text-column'
+            />
+        />
```
klezi10, line 172: I'm getting squiggly lines here but no error or any other kinds of messages about it. It seems to work though, and it doesn't work without it. Just wondering.

image

klezi10 (approved):
Looks good!

pages/about.js
cherylli: what error do you see if you mouse over on the error? I don't see any errors here in either vscode or webstorm

klezi10: When I mouse over, I don't see any error messages there either. Maybe it's my settings?

cherylli: oh I actually saw it briefly after i rebooted my computer and restarted vscode. It says loading, then after a while the red <TwoColumn is gone. Sorry no idea,

cherylli: I see it now. I'm supposed to wrap it in {} ...
image

mariana-caldas (approved):
Great job, @cherylli ! Thanks!
I've adjusted the `ButtonLink` component, so we can open links in an external tab and also changed the style to make sure it looks good on most desktop points, it wasn't great on 1024px before for example.

cherylli:
Do we need to change the text and link for "How to get started?" since we changed it so they are not sent to slack first?

mariana-caldas:
> Do we need to change the text and link for "How to get started?" since we changed it so they are not sent to slack first?

Great catch, @cherylli ! 🥇 
I've updated the content in the [mockup](https://www.figma.com/file/Fp0lLB5OWZLJseTs2svGbg/Web-Dev-Path_design?node-id=414%3A936) to:

"After taking a look at our project README and wiki, just send us an email sharing with us about your journey in tech and why you’re interested in joining us.
We've got you!"

I've added the links there to the README and wiki as well. The "Ping Us" link should take the users to the project email, as we see on the "Join us" navigation link. Please let me know if that makes sense, Cheryl! 🚀

cherylli:
updated. should be good to go

pages/about.js
```diff
@@ -127,6 +132,78 @@ export default function AboutUs() {
...
+        <TwoColumn
+          title='Wanna learn more?'
+          content={''}
+          image='/images/svg/close-curly-bracket.svg'
+          color={primary}
+          bgColor={lightBgColor}
+          customInnerClass='wanna-learn-more'
+        />
+        <TwoColumn
+          title={<div>Junior Developers</div>}
```
klezi10, line 145: Is it necessary to have the div wraps here? I think it also works with title='Junior Developers' or 'Experienced Developers'.

klezi10:
Just noticed that the "contact us" button under "experienced developers" looks different in the design
https://www.figma.com/file/Fp0lLB5OWZLJseTs2svGbg/Web-Dev-Path_design?node-id=414%3A936

Not sure if it was discussed elsewhere but thought I'd bring it up just in case.

pages/about.js
cherylli: No probably can be removed now. I had it so I can put a line break between the 2 words as in the design but Mariana removed it.

cherylli:
> Just noticed that the "contact us" button under "experienced developers" looks different in the design
> https://www.figma.com/file/Fp0lLB5OWZLJseTs2svGbg/Web-Dev-Path_design?node-id=414%3A936
> 
> Not sure if it was discussed elsewhere but thought I'd bring it up just in case.

I just assume it's to show the hover style 🤔. Might need a confirmation from the @Web-Dev-Path/designers

mariana-caldas (changes requested):
This is so great! Thanks so much, @cherylli !
There is only one thing I think we should adjust, please.
Only on mobile, can we remove the "/" and the section padding-bottom? 
image

mariana-caldas (approved):
LG(reat)TM! 🚀 👩🏼‍🎤

PR #133, opened 2022-09-06
Add contact us form (only visual)

klezi10:
| [**126**](https://github.com/MarianaSouza/web-dev-path/issues/126) |

Yes. 

#### What is this change?
Added the first section of Contact Us page which is the form and decorations. Also put a smaller width on the page's hero text content.

-contact us content width creates conflict on some screens to accommodate the bracket. It messes with the main page hero content width for example.

-buttonlink when added, I'm having trouble connecting it to the onsubmit function. I've added notes where the form button css starts and ends to easily find it. Currently the form console logs the data collected from the form.

-the form on position: absolute when scrolling, it floats over the navbar.

Locally.

After implementing functionality.


cherylli:
Is this what you mean?
image

I understand what you mean with ButtonLink component. At this stage, I'm not sure if it's possible to use it with handleSubmit without rewriting the ButtonLink component, as ButtonLink is built on <Link> and we don't really want to redirect with the submit. I will do more research and testing on this. 

Remove position: relative from Hero_header should work for the form going above nav bar. Also z-index: 10 above it is also not needed
image

image

The `ContactUsForm` component should not be in /pages otherwise users can access this page - http://localhost:3010/ContactUsForm (Next.js file-system based routing)

klezi10:
Thanks for your help @cherylli . I removed the z-index but left the position: relative because it stretched the hero background photo and changed it to this:
image

In any case it works now - the form doesn't float above the navbar.

klezi10:
> -contact us content width creates conflict on some screens to accommodate the bracket. It messes with the main page hero content width for example.

This is what I mean:
image

Unless there is another way, I think this is the best I can come up with but it changes the design a bit. I increased the width for the content and moved the bracket closer to the form for desktop screen size. Looks like this:
image
and
image

I made these changes:
image
image

mariana-caldas:
Hey @kles! Such a great job here! 🚀 
I see @cherylli is supporting with great ideas too! 🌟 

You're right about the use of a different button here since we need one with a type submit. In the near future, we'd probably need to transform it into a component that would be used on the subscribe to the newsletter as well.
I'm leaving some comments and more ideas to explore in terms of styling below. Please let me know your thoughts!

https://www.loom.com/share/b5c77c5dbdbf46a192cc4e17c2abe558

cherylli:
Thinking about the submit button. I think we shouldn't add it to the existing button link as it seems they are completely different. It will be too hard and messy to implement if possible at all

So, I think we have several options, 

1.  Just have it in the contact form 
`<button type='submit' className={contactUsFormStyles.contact__button}>Submit</button>`

2. wrap it with a container and style the container, possibly wrap the original "ButtonLink" so we can apply the same styles

image

3. a submit button component and add styles
image


in both 2 and 3 they are reusable and can customize the button text for example. I think I prefer #3 as it's simplier and more straightforward. But 2 looks good too if we can use it for both sumbit button and button link. 

What do you all think?

klezi10:
@cherylli Simpler and straightforward sounds good. Let's see what @mariana-caldas wants to do.

cherylli:
> Thanks for the feedback @mariana-caldas . I think I made the changes you mentioned in the video. On my end, I see this breaking from the homepage, just wanted to see if you see it on your side too @mariana-caldas and @cherylli . I had to make the width of the content in contact us smaller to allow for the bracket in some desktop screen sizes. If there is any other way to do it let me know, or make the changes and I'll take a look.
> image
> 

I was actually trying to reproduce that last night with the exact width and height, but not seeing that. Might try again tonight. I used Firefox

klezi10:
> I was actually trying to reproduce that last night with the exact width and height, but not seeing that. Might try again tonight. I used Firefox

I thought that might be the case! Let me know if get a chance to check again tonight.

cherylli:
Upon some investigation, this is due to the height rather than the width. It happens for a short time during scrolling.
image
image

Changing height to min-height here seems to work well for the frontpage 
image

But it still looks quite cramped on this height/width combination till you scroll down more
image

klezi10:
This PR is going to be replaced by the new one, [attached](https://github.com/Web-Dev-Path/web-dev-path/pull/134).

PR #145, opened 2022-11-17
Refactor CardColumns to Accept and array of card objects as props

cherylli:

yes

#### What is this change?
Refactor CardColumns to Accept and array of card objects as props

no

run it locally

after review

### DIFF
--- a/components/ContactUsCards.js
+++ b/components/ContactUsCards.js
@@ -2,11 +2,39 @@ import contactCardStyles from '@/styles/ContactUsCards.module.scss';
 import CardsColumns from './containers/CardsColumns';
 import RevealContentContainer from './containers/RevealContentContainer';
 
+const cards = [
+  {
+    title: 'FAQ',
+    image: '/images/svg/faq-icon.svg',
+    altTag: 'FAQ',
+    content: 'You can find the main questions about the project.',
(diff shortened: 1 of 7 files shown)


components/ContactUsCards.js
```diff
@@ -2,11 +2,39 @@ import contactCardStyles from '@/styles/ContactUsCards.module.scss';
...
+    linkText: 'Our Repository',
+  },
+];
+
 export default function ContactUsCards() {
   return (
     <div className={contactCardStyles.contactCards}>
       <RevealContentContainer>
         <CardsColumns
+          cards={cards}
```
mariana-caldas, line 37: Now we have the cards object here, can we remove lines 38-55, @cherylli ?

mariana-caldas (commented):
I love the idea of using an object prop `card` to store the card's data instead of delegating that responsibility to the `CardsColumns`, @cherylli , great catch!

Other than that comment above, I think this is good to go.

components/ContactUsCards.js
cherylli: yes, I forgot to remove them. thank you 🙂

mariana-caldas (approved):
Great refactor, @cherylli !

PR #146, opened 2022-11-18
Add LinkedIn link / Styles adjustments: newsletter button and contact

mariana-caldas:

Yes.

#### What is this change?

We're adding the LinkedIn link to the footer and also adjusting the newsletter button styles as well as the contact page height on desktop.

No.

Locally.

After review.

### DIFF
--- a/styles/SubmitButton.module.scss
+++ b/styles/SubmitButton.module.scss
@@ -40,8 +40,10 @@
       border: 1px solid $dark-bg-color;
     }
 
-    @include desktop {
+    @include large-desktop {
       min-width: 12rem;
+      border-radius: 3rem;
+      padding: 0.7rem 0;
     }
   }
 }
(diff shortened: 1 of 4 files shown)


styles/SubmitButton.module.scss
```diff
@@ -40,8 +40,10 @@
       border: 1px solid $dark-bg-color;
     }
 
-    @include desktop {
+    @include large-desktop {
```
cherylli, line 43: Not related to your changes but I see in the same file under .btn there's 2 "min-width"
image

cherylli (approved):

styles/SubmitButton.module.scss
mariana-caldas: The first media-query only refers to the .btn, @cherylli , while the second one only applies to .newsletter__button, so there is no redundancy here. Please let me know if that is what you were talking about, Thanks 🙌🏼 

image

image

cherylli: I mean line 6 and line 14

mariana-caldas: Oh, I see now, @cherylli ! My bad!
Making a note to fix that next PR. I will make a PR release to 1.2.0 in the next few days before we start thinking about the blog.

PR #147, opened 2022-11-27
Release 1.2.0

mariana-caldas:

Yes.

#### What is this change?

* Release 1.2.0, which means we're tagging on GitHub all the features and fixes we've been doing since 1.1.0
* Upgrading Next.js to 13

Yes.

To upgrade to Next 13, I had to perform some adjustments to the site referent to the `Link` and `Image` components.

https://nextjs.org/docs/upgrading
https://nextjs.org/docs/messages/invalid-new-link-with-extra-anchor
https://nextjs.org/docs/messages/next-image-upgrade-to-13

Locally.

After review. 
_**Please run yarn locally to upgrade your local copy with Next 13 before testing the branch**_


components/containers/Card.js
```diff
@@ -11,7 +11,9 @@ export default function Card({
       {linkText}
     </a>
   ) : (
-    <Link href={link}>{linkText}</Link>
+    <Link href={link} legacyBehavior>
```
cherylli, line 14: I think this is only needed if the child is <a>
image

mariana-caldas: Good catch! Thanks, @cherylli !
The curious thing here is the property was added automatically when I run the command suggested on [their documentation](https://nextjs.org/docs/messages/invalid-new-link-with-extra-anchor), however, now I remove it, there is no error, so removed for now 🙌🏼

cherylli (approved):

klezi10 (approved):

PR #154, opened 2023-01-19
Adjusting blog container and titles styling

mariana-caldas:

Yes.

#### What is this change?

We're adjusting the blog container and its titles styling

No

Locally.

_Before:_

image

image

<br />

_After:_

image

image

After review

### DIFF
--- a/styles/Blog.module.scss
+++ b/styles/Blog.module.scss
@@ -10,12 +10,14 @@
 .viewAll {
   font-size: 1.5rem;
   font-weight: bold;
-  margin: -6rem 6.5rem 4rem 0;
-  text-align: right;
+  position: relative;
+  top: -3rem;
+  display: block;
+  text-align: center;
 
-  @include mobile {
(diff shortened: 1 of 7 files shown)


styles/Blog.module.scss
```diff
@@ -4,24 +4,32 @@
 .blogContainer {
   display: flex;
   flex-direction: column;
+  margin: 3rem 0;
```
klezi10, line 7: Just wondering if it would be shorter to do
`margin: 0.5rem 0 3rem;`

then remove the margin-bottom on line 28 to avoid using negative values?

Though having both margin properties seems easier to read? @mariana-caldas

mariana-caldas: > 

We may need to refactor those margins indeed when we uncomment the blog search whose negative margin is applied indeed, @klezi10 . Thanks for noticing that!

klezi10 (approved):

Issue #155, opened 2023-01-20
Implement Search Bar functionality 


mariana-caldas:
**What do we need to build or fix?**

Currently, we have the [front-end search functionality](https://github.com/Web-Dev-Path/web-dev-path/blob/main/components/blog/SearchBar.js) and it is commented out on the [blog page](https://github.com/Web-Dev-Path/web-dev-path/blob/main/pages/blog/index.js#L18), we need to add the logic so that any user can search the blog posts data on that page.

**Technical details**
- Implement the functionality by considering the blog posts

**Approach suggestions**
 - On the blog `index.js`,  add the `revalidate` prop to the `getStaticProps` function to use the [Incremental Static Generation Approach
](https://nextjs.org/docs/basic-features/data-fetching/incremental-static-regeneration)
- Then, by considering the `posts` object returned from the `getStaticProps` function, add the search logic. Suggested tutorials about the search logic implementation:
   https://www.youtube.com/watch?v=ZoayCCDHFiI / https://www.youtube.com/watch?v=lOIrlONj3So

**Deadline**

Please keep in mind that once you assign this task to yourself, you'll need to complete it in 20 days.

**Acceptance criteria**

- Test the section and components in many screen sizes, you can use the Inspect tool for that.
- Please test if the new changes added to the components do not affect the other instances.
- Test the feature in many browsers, such as Chrome, Firefox, Edge, and Safari (MAC).
- If there are any build problems when submitting your PR, run `yarn build` locally to solve the issues and commit the changes.
- Update the CHANGELOG.md file.

cherylli:
I don't think the link is relavent is it's for a next.js "markdown blog" which is not what we have now. It has articles in the file system and we don't. We already have all the posts fetched (using SSR which I think we should switch to ISR), so we could just search in that posts array and display it using the BlogContainer component

mariana-caldas:
> @mariana-caldas I don't think the link is relavent is it's for a next.js "markdown blog" which is not what we have now. It has articles in the file system and we don't. We already have all the posts fetched (using SSR which I think we should switch to ISR), so we could just search in that posts array and display it using the BlogContainer component

Great point, @cherylli , thanks! I've updated the issue with your idea 🥳 . However, just to clarify for everyone reading this is the future YOU DON'T have to follow anything suggested under "approach suggestions"; you may have a better way to tackle things so never limit yourself!

cherylli:
[...]

PR #156, opened 2023-01-31
Feature/blog search

cherylli:
| [**155**](https://github.com/MarianaSouza/web-dev-path/issues/155) |
Closing #155
yes

#### What is this change?
add search bar funcationality in blog page

no

run locally

after review

### DIFF
--- a/pages/blog/index.js
+++ b/pages/blog/index.js
@@ -5,17 +5,18 @@ import Title from '@/components/snippets/Title';
 import styles from '@/styles/Blog.module.scss';
 import { blogRevalidate } from '@/utils/config';
 import { tagToHeading } from '@/utils/blogCategories';
+import { blogSearch } from '@/utils/search';
 
 export default function Blog({ posts }) {
   const [searchResults, setSearchResults] = useState(null);
-  const [searchTerm, setSearchTerm] = useState('temp');
+  const [searchTerm, setSearchTerm] = useState('');
   const getPostsByTag = tag => {
     return posts.filter(post => post.tagList.includes(tag));
   };
 
   useEffect(() => {
-    console.log('search result', searchResults);
-  }, [searchResults]);
+    setSearchResults(blogSearch(posts, searchTerm));
+  }, [searchTerm]);
 
   return (
     <>
--- a/pages/blog/index.js
+++ b/pages/blog/index.js
@@ -25,7 +25,7 @@ export default function Blog({ posts }) {
           title={searchResults ? '' : 'Latest Posts'}
         />
         <SearchBar
-          posts={posts}
+          items={posts}
           setSearchTerm={setSearchTerm}
           setSearchResults={setSearchResults}
         />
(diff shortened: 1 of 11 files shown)


mariana-caldas (approved):
Great job, @cherylli !

I've refactored the implementation on the blog page to make it more readable. I'd also suggest trying to replace some of the stylings we have for wrapping the components into `<Container />` if possible because that component already contains many of the stylings we're repeating -> https://github.com/Web-Dev-Path/web-dev-path/blob/main/styles/Container.module.scss

cherylli:
> Great job, @cherylli !
> 
> I've refactored the implementation on the blog page to make it more readable. I'd also suggest trying to replace some of the stylings we have for wrapping the components into `<Container />` if possible because that component already contains many of the stylings we're repeating -> https://github.com/Web-Dev-Path/web-dev-path/blob/main/styles/Container.module.scss

but it will affect everything which uses <Container /> right? unless we add a custom class
but if we are looking at changing over to styled components maybe we can look at cleaning it up when we do that?

cherylli:
just formatted the search results phrase (results=> result/results, 0 search result => no search result)
and removed "view all" from search result as it's should already viewing all

Faith-Nchifor:
Thumbs up @cherylli. Great Job

mariana-caldas:
> > Great job, @cherylli !
> > I've refactored the implementation on the blog page to make it more readable. I'd also suggest trying to replace some of the stylings we have for wrapping the components into `<Container />` if possible because that component already contains many of the stylings we're repeating -> https://github.com/Web-Dev-Path/web-dev-path/blob/main/styles/Container.module.scss
> 
> but it will affect everything which uses right? unless we add a custom class but if we are looking at changing over to styled components maybe we can look at cleaning it up when we do that?

Good thinking, @cherylli ! I'm making a note so we circle it back when we implement the styled-components!

pages/blog/index.js
```diff
@@ -1,25 +1,45 @@
...
+    posts: posts.slice(0, 3),
+  };
+
+  if (searchTerm) {
+    const filteredPosts = blogSearch(posts, searchTerm);
+    filteredData.posts = filteredPosts;
+    filteredData.heading = `${
+      filteredPosts.length === 0 ? 'no' : filteredPosts.length
+    } search ${
+      filteredPosts.length > 1 ? 'results' : 'result'
```
klezi10, line 23: Nice! Interesting way of setting this up!

utils/config.js
```diff
@@ -0,0 +1 @@
+export const blogRevalidate = 60 * 60 * 24;
```
klezi10, line 1: I understand `export const blogRevalidate = 60`; Nextjs will try to re-generate the page every 60 seconds. 

What does `60 * 60 * 24` mean? Is it 8460 seconds which is about 2.5 hours? If so, wondering why set it up that way? 

I did a few Google searches but all the examples came up with either 10 seconds or 60 seconds.

klezi10 (approved):

utils/config.js
cherylli: 60s x 60s/h = 3600s/h
3600s/h x 24h/day = 86400s/day
(60s x 60s/h x 24h/day = 86400s/day)

It's common to write it this way so it's easier to see compared to 86400

So it sets the revalidation time to 1 day which means if the last request is older than one day it will refetch in the background.

It's just an arbitrary value that I think is reasonable, but this value can be adjusted base on post update frequency in future.

PR #161, opened 2023-02-15
Added styled-components

klezi10:
| [**159**](https://github.com/MarianaSouza/web-dev-path/issues/159) |

Yes

#### What is this change?
Upgraded Nextjs, installed styled-components, applied styled-components to `Title` component.

No

Locally

After review


klezi10:
Good morning :) The PR is in draft mode as I just noticed I need to modify the `blogPostsContainer` title.

components/snippets/Title.js
```diff
@@ -1,15 +1,19 @@
...
-    >
-      {title}
-    </h2>
-  );
+export const TitleStyle = styled.h2`
+  ${props =>
+    props.blogTitle &&
+    `
+    align-self: flex-start;
+    margin: ${props.adjustedTitle ? '0 auto 1rem 15rem' : '0 auto 1rem'};
```
klezi10, line 8: I'm not entirely sure what's going on but for some reason this works for the blog post container titles (like Next.js and Typescript titles" but I didn't even pass the prop. If I remove the adjusted Title prop at all, the titles are off margin. Wondering what it looks like from your end @mariana-caldas @cherylli

styles/Blog.module.scss
```diff
@@ -23,15 +23,14 @@
 
 .blogSearch {
   display: flex;
-  align-items: center;
   padding-top: 2.5rem;
-  margin: 0 auto;
-  width: 90%;
-  max-width: $large-desktop-breakpoint;
+  justify-content: space-between;
```
mariana-caldas, line 27: A best practice to code web products really responsive, @klezi10 , is to start coding having the mobile in mind, and then add a media query for the desktop case. That's the mobile-first approach: 
https://www.wix.com/blog/2022/07/mobile-first-design/?utm_source=google&utm_medium=cpc&utm_campaign=19399843409^145175008536&experiment_id=^^642983295314^^_DSA&gclid=Cj0KCQiArsefBhCbARIsAP98hXRkphy6wod_HH--iD5L3NUifvZ71dmdxbO5piEJF81LvohoYDmhbFkaAnCaEALw_wcB

mariana-caldas (approved):
Way to go, @klezi10 ! 🚀 
Please let me know if my comments make sense.

components/snippets/Title.js
```diff
@@ -1,15 +1,12 @@
...
-      className={
-        customClass ? `${styles.title} ${styles[customClass]}` : styles.title
-      }
-    >
-      {title}
-    </h2>
-  );
+export const StyledTitle = styled.h2`
+  align-self: flex-start;
+  margin: '0 auto 1rem';
```
cherylli, line 5: This line gives me an error because of the single quotes it is parsed as string. 
`margin: 0 auto 1rem` is showing as valid and applied but the title position moved to the center of the parent div which seems to be exactly what this line is doing,

removing the whole line works fine for me locally. 

Could you please confirm @klezi10 because marianna and I are seeing different things

invalid
image

valid (after removing the single quotes)
image

removing the line shows same as "invalid" as it wasn't applied

cherylli (approved):

PR #163, opened 2023-02-25
Fix Card container tag list on blog page overflow when there are 3 or more tags

cherylli:

Yes

#### What is this change?
Fix the overflowing issues of blog card tag list when there are 3 more tags
Also limit the max number of tags shown to 8 (which would be about 4 rows for longer tags)
image

image

no

run it locally

after review


mariana-caldas (approved):
That's a great catch, @cherylli ! You're the best, thanks! <3

PR #164, opened 2023-03-11
Add Links to Blog Tags

cherylli:

yes

#### What is this change?
- [x] Add links to blog tags so it displays all posts with the same tag (/blog/category/tag)
- [x] Add back button so users can go back to the main blog page from the tag page

When there's not title match for tags (in utils/blogCategories), the tag (#tag) will show as title.

no but I had to use !important in the link text-decoration as the tags are inside card so inherited the card link styles and couldn't be overwritten

#### When should this be merged? After review

image

### DIFF
--- a/components/containers/Card.js
+++ b/components/containers/Card.js
@@ -1,7 +1,7 @@
 import Image from 'next/image';
 import Link from 'next/link';
 import styles from '@/styles/Card.module.scss';
-import Tag from '@/components/blog/Tag';
+import { Tag, TagContainer } from '@/components/blog/Tag';
 
 export default function Card({
   card: { image, altTag, title, content, link, linkText, tagList },
@@ -30,11 +30,11 @@ export default function Card({
         {title}
       </h2>
       {tagList && tagList.length > 0 ? (
-        <div className={styles.tagListContainer}>
+        <TagContainer>
           {tagList.slice(0, 8).map((tag, i) => (
             <Tag key={i} text={tag} />
           ))}
-        </div>
+        </TagContainer>
       ) : null}
       <div className={styles.content}>
         <p>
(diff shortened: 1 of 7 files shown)


mariana-caldas (approved):
LGTM, @cherylli ! Thanks!

I used `styled-components` to convert `Tag` so we don't need to style it on `Blog` anymore, which led to using `! important` to overwrite `Card` styles. Please let me know your thoughts!

components/containers/Card.js
```diff
@@ -1,7 +1,7 @@
 import Image from 'next/image';
 import Link from 'next/link';
 import styles from '@/styles/Card.module.scss';
-import Tag from '@/components/blog/Tag';
+import { Tag, TagContainer } from '@/components/blog/Tag';
```
cherylli, line 4: Why not contain TagContainer in the Tag component so we won't need the import? Also keeps everything tag related inside Tag

mariana-caldas: Because we only need to render one container for many `Tag` components on the map. Have you tried it locally, @cherylli ?

cherylli: I misunderstood it. Please ignore my comment. I've never thought of importing styles from a child component

mariana-caldas: It's super important to ask questions, Cheryl! The beauty of working in a team with version control is the possibility of learning from each other. 
My only suggestion is to always try it locally too to ensure it really makes sense to you.
Yeah, building styles on the fly is so cool! And that's why styled-components will kill our customClass props. My idea should be refactored later tough since, ideally, the styles should be in a different file -> https://blog.cloudboost.io/separate-your-code-with-styled-components-ec4fd1ee3ef8

Issue #162, opened 2023-02-20
Implement `styled-components` to the project


mariana-caldas:
**What do we need to build or fix?**

We need to implement [styled-components](https://styled-components.com/docs/basics) to the project  by converting all the components to use this new library for styling.

**Technical details**

- Create a new branch named `feature/styled-components-conversion` and checkout to it.
- Update all the components to use `styled-components` for styling.
- Remove any existing inline styles or global styles that are not necessary.
- Test the components in different screen sizes and browsers to ensure they look and function as expected.
- Commit the changes to the `feature/styled-components-conversion` branch and create a pull request to merge the changes into the `main` branch.
- Follow the discussion about the new styled-components name convention on issue #160 

**Approach suggestions**

- https://styled-components.com/docs/basics#getting-started

**Deadline**

Please keep in mind that once you assign this task to yourself, you'll need to complete it in 3 weeks.

**Acceptance criteria**

- All components should be styled using styled-components.
- All components should look and function as expected in different screen sizes (you can use the Inspect tool for that) and browsers(Chrome, Firefox, Edge, and Safari for MAC)
- If there are any build problems when submitting your PR, run `yarn build` locally to solve the issues and commit the changes.
- Update the CHANGELOG.md file.

PR #166, opened 2023-03-30
Styled components conversion

thorrellt:
| [**162**](https://github.com/Web-Dev-Path/web-dev-path/issues/162) |

yes

#### What is this change?
Implement styled-components to the project by converted all the components to use them. 

yes, including:  

- handling the html loading in before the styles by changing the _document.js file.
- handling the different style variations of certain components, especially TwoColumns.
- removing dependencies on className and instead using style attributes.

in my local Firefox and Chrome browser

This is ready to be merged immediately after being reviewed.


mariana-caldas:
Good job on submitting a DRAFT so your decisions become more visual, @thorrellt !

I think your idea of grouping the buttons altogether is on point, however, we're adding `styled-components` to the project now to remove the `customClasses` that make the styling so confusing whenever possible. So, my only suggestion, for now, would be to create more buttons that extend `ButtonLink` functionalities while having their proper CSS, such us `ButtonLinkInvertedGrey` for example. 

Please let me know if that makes sense, Thorrell. If not, don't sweat! I can commit to your branch explaining that better 🙌🏼

thorrellt:
I understand. Thanks for the feedback. I'll update this component and update the rest with this in mind.

thorrellt:
I started reworking ButtonLink this morning, and kinda got stuck. It was recommended that I extend ButtonLink to make more buttons ie: ButtonLinkInvertedGrey. My sticking point is that ButtonLinkInvertedGrey is/would never be specifically invoked. The color scheme is passed as a prop from index -> TwoColumn -> ButtonLink. If I start extending styled component in this case, I feel like I would have to:

A. Put some kind of switch or if/else login in TwoColumn to read the props and return each specific button.
B. And by that logic, I should actually even extend TwoColumn and make multiples of it to be called by index.js and wherever it is needed (I haven't looked. Only worked in a small section of code as of yet). 

I still think using props & transient props would be appropriate. This would still remove the dependency on classNames and remove excess classes from our final HTML, although our code would look only slightly different (since attributes and props are written in the exact same way.) I can think of use cases for extending a component but in this case, leveraging props seems the most straightforward, not to mention how much the code base would grow.

However, I am sure I am ignorant to a lot of factors. Am I overlooking something? or, is there a more elegant solution that would allow me to extend components without adding a lot of extra switches and code?

Any thoughts?

mariana-caldas:
> I started reworking ButtonLink this morning, and kinda got stuck. It was recommended that I extend ButtonLink to make more buttons ie: ButtonLinkInvertedGrey. My sticking point is that ButtonLinkInvertedGrey is/would never be specifically invoked. The color scheme is passed as a prop from index -> TwoColumn -> ButtonLink. If I start extending styled component in this case, I feel like I would have to:
> 
> A. Put some kind of switch or if/else login in TwoColumn to read the props and return each specific button. B. And by that logic, I should actually even extend TwoColumn and make multiples of it to be called by index.js and wherever it is needed (I haven't looked. Only worked in a small section of code as of yet).
> 
> I still think using props & transient props would be appropriate. This would still remove the dependency on classNames and remove excess classes from our final HTML, although our code would look only slightly different (since attributes and props are written in the exact same way.) I can think of use cases for extending a component but in this case, leveraging props seems the most straightforward, not to mention how much the code base would grow.
> 
> However, I am sure I am ignorant to a lot of factors. Am I overlooking something? or, is there a more elegant solution that would allow me to extend components without adding a lot of extra switches and code?
> 
> Any thoughts?

Thanks for communicating so well your process, @thorrellt ! That's the spirit!
I'm taking a look at your branch this weekend to test the possibilities with you in the next few days. Have a great weekend!

mariana-caldas:
By taking a look at the `TwoColumn` component (attached) I see what you mean, @thorrellt . We can extend the `ButtonLink` to other custom possibilities to get rid of the `props`, however, adding conditionals there would make the code less readable on `TwoColumn` and I agree with you that it does not worth it. We may be able to follow the extended approach without `props` and `customClass` for the others, let's see... 

image

Good catch on creating a JS file for variables and mixins! 🥇  Before we forget it, I changed a hard-coded color on the `Tag` component.  Please don't hesitate to reach out here or on Slack anytime you feel stuck.

components/containers/Container/Container.js
```diff
@@ -1,14 +1,15 @@
 import containerStyles from '@/styles/Container.module.scss';
 import S from './styles';
 
-export default function Container({ customClass, children, styles }) {
+export default function Container({
+  className,
```
mariana-caldas, line 5: Hey, @thorrellt !
Can you check if we still need all those props here? We may be able to keep only "children" here if the other instances allow us to extend the `Container`, for example.

thorrellt: Okay. I was going to check props once I was done with all the styled props. I try to remove dead code when I find it, and I plan to run through everything before I submit my final pull. ...consider it done

thorrellt:
The TwoColumns component was very dependent on custom classes. There are a lot of variations of it's components based on these classes.  In order to keep the files kinda clean and readable I put all that logic in a separate file, and return the css via a function. like: 
image

If anyone has the time, can they check out my thought process and let me know if it makes sense? and if they can think of a better solution if this isnt a good design choice? thanks

mariana-caldas:
> The TwoColumns component was very dependent on custom classes. There are a lot of variations of it's components based on these classes. In order to keep the files kinda clean and readable I put all that logic in a separate file, and return the css via a function. like: image
> 
> If anyone has the time, can they check out my thought process and let me know if it makes sense? and if they can think of a better solution if this isnt a good design choice? thanks

Hi @thorrellt !

I totally see your point about the `addContentsCss.js` file and the number of props we have for the `TwoColumn`. What if we try to simplify it by having 3 TwoColumn components? The first would be the regular `TwoColumn.js` with no custom classes. The second and third would extend TwoColumn to become TwoColumnHomepage and TwoColumnAboutUs with all their respective props.

I suggest that because the approach with addContentsCss seems less readable to me than the regular Sass approach we were using before for that confusing component, so maybe that's a repetition that would make us understand better the codebase. How do you see that?

thorrellt:
> Hi @thorrellt !
> 
> I totally see your point about the `addContentsCss.js` file and the number of props we have for the `TwoColumn`. What if we try to simplify it by having 3 TwoColumn components? The first would be the regular `TwoColumn.js` with no custom classes. The second and third would extend TwoColumn to become TwoColumnHomepage and TwoColumnAboutUs with all their respective props.
> 
> I suggest that because the approach with addContentsCss seems less readable to me than the regular Sass approach we were using before for that confusing component, so maybe that's a repetition that would make us understand better the codebase. How do you see that?

I see where you're coming from. I'm not sure extension is possible, as it seems to be tricky with components, especially with nested components because the extension doesn't account for them. However, doing two different implementations could/should work. 

I will play with it and do some research over the next few days. Either way, there will be multiple TwoColumn components to improve readability.

components/buttons/SubmitButton/styles.js
```diff
@@ -0,0 +1,73 @@
...
+    //check props for button variations
+    ${props =>
+      props.$buttonType === 'newsletter__button' ? NewsLetterButton : ''}
+  `)}
+
+  //media query mixins
+  ${m.largeDesktop(css`
+    //check props for button variations
+    ${props =>
+      props.$buttonType === 'newsletter__button'
```
mariana-caldas, line 44: To make the code more readable, I'd rather remove the `buttonType` prop and extend the `SubmitButton` to the other ones if possible.

components/containers/Card/styles.js
```diff
@@ -0,0 +1,197 @@
...
+    margin: 1.5rem 1.5rem 0 1.5rem;
+    &:first-child,
+    &:last-child {
+      margin: 1.5rem 0.5rem 0 0.5rem;
+      ${props => (props.$cardType === 'blog' ? Blog : '')}
+    }
+  `)}
+
+  //check props for button variations
+  ${props => (props.$cardType === 'our-goals' ? AboutUs : '')}
```
mariana-caldas, line 51: I'd rather see the Card component extended to other "children" than props.

mariana-caldas (changes requested):
You're are doing a great job so far, @thorrellt !

I'd like to ask for some changes about the first components I could review and I ask you to reflect on them for the other components' conversion before I go forward, please.
The main reason we're converting those components into `styled-components` is to improve the readability so I'd rather see more extended components than `props` or custom classes whenever possible, please.

cherylli (commented):
Looks good. Thanks for doing this, it's a very big task!

Not sure if it's just me, is anyone else getting this on every page load?
image

mariana-caldas:
> Looks good. Thanks for doing this, it's a very big task!
> 
> Not sure if it's just me, is anyone else getting this on every page load? image

Hey, @cherylli ! Thanks for your feedback. 
I don't see those errors when I run this branch locally from a Windows PC on Chrome (attached). However, I see some warnings referent to our `pwa` and `styled-components` that may require we update our Next.js version to the latest and, then, change a little bit our `next,config.js` file later in another task. Please let us know if you need some supporting to find out why do you see those errors in your end, I'd recommend to remove your `node_modules`  and install the dependencies again just to make sure any weird package is corrupting your local environment.

image

cherylli:
> > Looks good. Thanks for doing this, it's a very big task!
> > Not sure if it's just me, is anyone else getting this on every page load? image
> 
> Hey, @cherylli ! Thanks for your feedback. I don't see those errors when I run this branch locally from a Windows PC on Chrome (attached). However, I see some warnings referent to our `pwa` and `styled-components` that may require we update our Next.js version to the latest and, then, change a little bit our `next.config.js` file later in another task. Please let us know if you need support to find out why you see those errors on your end, I'd recommend removing your `node_modules` and installing the dependencies again to make sure any weird package is corrupting your local environment.
> 
> image

I've deleted .next and node_modules and still seeing this, but not in the terminal. I saw it in the browser inspector. It seems to be only a chrome issue for me. I might look into it when I have more time.

thorrellt:
> You're are doing a great job so far, @thorrellt !
> 
> I'd like to ask for some changes about the first components I could review and I ask you to reflect on them for the other components' conversion before I go forward, please. The main reason we're converting those components into `styled-components` is to improve the readability so I'd rather see more extended components than `props` or custom classes whenever possible, please.

So can you look at how I did Two Columns and let me know if that is acceptable?  Two columns has 11 variations that are made up of probably 20+ extended components. I packaged these components into objects then assign them in TwoColumns.js based on a a singled  prop ($contentType) that is received when TwoColumns is called. It is passed and referenced only one time. This way when "about-us" references the <S.Title> component, S.Title is a key that actually references S.AboutUsTitle. 

If I extend TwoColums then we will have 11 versions of it b/c I cannot think of how else to change the internal components that make up the TwoColumn container. If ever any changes need to happen to it, it will have to be repeated 11 times.

Do you think this design pattern will work? It 
- eliminated the use of props to inject CSS like u referenced before.
- removed all props passing from inside the parent container to clean up the code.
- All the logic for which set of extended components to use is done in these lines of code
image

Thoughts?

CHANGELOG.md
```diff
@@ -78,6 +79,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - next.js warning - no stylesheets in head component
 - CardColumns refactoring to accept an array of card objects as props
 - styles on newsletter button and contact page
+- error where html loaded before styles by updating \_document.js
```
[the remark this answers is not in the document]
thorrellt: will do

mariana-caldas:
> > You're are doing a great job so far, @thorrellt !
> > I'd like to ask for some changes about the first components I could review and I ask you to reflect on them for the other components' conversion before I go forward, please. The main reason we're converting those components into `styled-components` is to improve the readability so I'd rather see more extended components than `props` or custom classes whenever possible, please.
> 
> So can you look at how I did Two Columns and let me know if that is acceptable? Two columns has 11 variations that are made up of probably 20+ extended components. I packaged these components into objects then assign them in TwoColumns.js based on a a singled prop ($contentType) that is received when TwoColumns is called. It is passed and referenced only one time. This way when "about-us" references the <S.Title> component, S.Title is a key that actually references S.AboutUsTitle.
> 
> If I extend TwoColums then we will have 11 versions of it b/c I cannot think of how else to change the internal components that make up the TwoColumn container. If ever any changes need to happen to it, it will have to be repeated 11 times.
> 
> Do you think this design pattern will work? It
> 
> * eliminated the use of props to inject CSS like u referenced before.
> * removed all props passing from inside the parent container to clean up the code.
> * All the logic for which set of extended components to use is done in these lines of code
>   image
> 
> Thoughts?

Yeah, I'd say to focus on revisiting the other components to check if we can extend them instead of using customClasses and leaving the TwoColumns as the least one while we think if there is another way to simplify it or if that's the best approach, @thorrellt . Please let me know if that makes sense!

thorrellt:
> > > You're are doing a great job so far, @thorrellt !
> > > I'd like to ask for some changes about the first components I could review and I ask you to reflect on them for the other components' conversion before I go forward, please. The main reason we're converting those components into `styled-components` is to improve the readability so I'd rather see more extended components than `props` or custom classes whenever possible, please.
> > 
> > 
> > So can you look at how I did Two Columns and let me know if that is acceptable? Two columns has 11 variations that are made up of probably 20+ extended components. I packaged these components into objects then assign them in TwoColumns.js based on a a singled prop ($contentType) that is received when TwoColumns is called. It is passed and referenced only one time. This way when "about-us" references the <S.Title> component, S.Title is a key that actually references S.AboutUsTitle.
> > If I extend TwoColums then we will have 11 versions of it b/c I cannot think of how else to change the internal components that make up the TwoColumn container. If ever any changes need to happen to it, it will have to be repeated 11 times.
> > Do you think this design pattern will work? It
> > 
> > * eliminated the use of props to inject CSS like u referenced before.
> > * removed all props passing from inside the parent container to clean up the code.
> > * All the logic for which set of extended components to use is done in these lines of code
> >   image
> > 
> > Thoughts?
> 
> Yeah, I'd say to focus on revisiting the other components to check if we can extend them instead of using customClasses and leaving the TwoColumns as the least one while we think if there is another way to simplify it or if that's the best approach, @thorrellt . Please let me know if that makes sense!

I understand @mariana-caldas. Much thanks for the feedback, and sorry for my slow response/follow-up.

components/containers/Container/Container.js
thorrellt: I removed the customClass, but the classname has to stay. It's part of extending Styled Components (i think.) I specifically added this as a fix a while back, so I forgot where is the docs I read this. But if we remove customer class, certain styles wont pass correctly.

components/layout/Footer/styles.js
```diff
@@ -0,0 +1,129 @@
...
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    text-align: left;
+  `)}
+`;
+
+const Logo = styled.a`
+  display: block;
+  max-width: 12.5rem;
```
mariana-caldas, line 27: Please remove the properties `display`, `max-width`, and `margin` here. They are preventing centering the logo properly on mobile.

components/mailchimp/NewsletterForm/styles.js
```diff
@@ -0,0 +1,224 @@
...
+  InputName,
+  Input,
+  Form,
+  FormError,
+  FormInfo,
+  FormSending,
+  FormSuccess,
+};
+
+// .newsletterFormInfo {
```
mariana-caldas, line 163: Please remove all the comments.

components/containers/Card/styles.js
```diff
@@ -0,0 +1,258 @@
...
+
+const BlogImageWrapperCss = css`
+  height: 12rem;
+`;
+
+const BlogImageWrapper = styled(ImageWrapper)`
+  height: 12rem;
+`;
+
+// export default {
```
mariana-caldas, line 221: Please remove the comments.

components/containers/CardColumns/CardColumnsInterface.js
```diff
@@ -1,24 +1,17 @@
...
 import 'swiper/css';
 import 'swiper/css/pagination';
 import 'swiper/css/navigation';
-import Card from '@/components/containers/Card';
 import Container from '@/components/containers/Container';
-import styles from '@/styles/CardsColumns.module.scss';
+import * as S from './styles';
 
-export default function CardsColumns({ cards, customClass }) {
+export function CardsColumnsInterface({ cards, customClass, cardComponent }) {
```
mariana-caldas, line 9: What a great idea to create an interface handled through a property!

mariana-caldas (changes requested):
We're almost there, @thorrellt ! Such a great job! 🥷

cherylli (commented):
Thanks for the changes. I didn't check all of the files but there are a few files with unused imports which should be removed. For example, in Container/rows/index.js and TwoColumn/index.js

the grey part are unused variables and imports
image
image

cherylli (commented):
Thanks for the new commit, I still see some unused imports and variables. Were they just missed or left there on purpose?

Otherwise everything looks good.

cherylli:
> 

My IDE (webstorm) shows unused imports/variables in grey, vscode has similar thing, I guess a lot of editors and IDE will have similar features

For example here customClass is not in used, you have removed some in this file but not customClass so I was just wonder if there's any other reasons for this

image

TwoColumns still has some unused imports 
image


But just go into each file and have a quick look.

webstorm does have an "Optimize Imports" function to remove all unused imports

This is also in Webstorm, you can scan for problems in the whole project.
image

components/containers/Card/CardInterface.js
```diff
@@ -0,0 +1,39 @@
+import { Tag } from '@/components/blog/Tag';
+import Link from 'next/link';
+
+export function CardInterface({
```
mariana-caldas, line 4: It makes total sense to see this component as an interface!

mariana-caldas (approved):
Such a great job, @thorrellt ! 🥇

cherylli (approved):
looks good!

Issue #165, opened 2023-03-16
Add theming to the project


mariana-caldas:
**What do we need to build or fix?**

Now we're using `styled-components` implemented on issue #162 , we need to add a theme to the project and keep all the styling variables in a single location to enable future reuse of the project as a theme base.

**Technical details**

- Create a new branch named `feature/add-theming` and checkout to it.
- Create a theme file named `themeConfig.js` in the root directory.
- Add the `createGlobalStyle` object to export the globalStyles per `theme`.
- Create a `lightTheme` object that will be exported and add the current styling variables to it.
- Import the `theme` file on `_app.js`.
- Leverage the `ThemeProvider` wrapper component provided by the styled-components library to implement the theme.
- Ensure all components use the variables that are transferred to the new theme system.
- Test the feature in different screen sizes and browsers to ensure it looks and functions as expected.
- Commit the changes to the `feature/add-theming` branch and create a pull request to merge the changes into the `main` branch.
- Update the `CHANGELOG.md` file.

**Approach suggestions**

- https://styled-components.com/docs/advanced#theming
- https://blog.logrocket.com/theming-in-next-js-with-styled-components-and-usedarkmode/

**Deadline**
Please remember that once you assign this task to yourself, you'll need to complete it in 3 weeks.

**Acceptance criteria**

- Test the section and components in many screen sizes, you can use the Inspect tool for that.
- Please test if the new changes added to the components do not affect the other instances.
- Test the feature in many browsers, such as Chrome, Firefox, Edge, and Safari (MAC).
- If there are any build problems when submitting your PR, run `yarn build` locally to solve the issues and commit the changes.
- Update the CHANGELOG.md file.

PR #175, opened 2023-06-14
add/typescript files

vmcodes:
| [174](https://github.com/MarianaSouza/web-dev-path/issues/174) |

Not yet, was looking for guidance.

#### What is this change?
Updated file names to be compatible with TypeScript and added TypeScript dependencies.

I think that it may be a good idea to at least change the files names before getting started, so that if any other issues are worked on at the same time, things won't get too messy when merging.

No.

Everything worked locally for me.

This should be reviewed first, beginning with a draft.


vmcodes:
I wanted to draft this PR just to get the discussion started around beginning to at least rename the files, so that other issues can still be worked on. If you all think this is a good idea, I can correct some of the issues seen in the build. Thank you.

cherylli (changes requested):
There are build errors. Don't think you can convert to ts just by changing the filenames. It won't build, you have to add types.

It might work locally for dev but it won't build for production.

cherylli:
I think we can keep the discussion in the issues thread. I thought you wanted reviews to merge into main since this is a pull request.

vmcodes:
@cherylli I did want to merge this to avoid issues later. I just wasn't sure if that would be a good approach.

cherylli:
Yeah I'm not sure. I don't think we should push any temporary thing to production unless it's an urgent bandaid fix.

vmcodes:
That's not an issue with me at all, do you think then maybe this issue should be completed before [165](https://github.com/Web-Dev-Path/web-dev-path/issues/165) since no one has started working on it yet? If so, I can just close this out.

mariana-caldas:
Hey, @vmcodes !

It seems like a nice start PR! The idea here was just to change the file naming and, then, keep doing the conversion file by file, right? If so, that's a way of doing it, but I have to agree with Cheryl: we don't push into production anything that is not fully ready because it's a risk you don't need to take.

We could create a `dev` branch to handle this approach. In that case, we would:

1. Create a `dev` branch where to keep merging the TS conversion increments. That would allow having more than one developer working with the feature if you get busy or stuck.
2. When the `dev` branch is ready - no questions nor build errors - we merge it into `main`, which will lead to prod.

How do you find this approach?

vmcodes:
That makes sense to me, I already changed all the file names though on chore/add-typescript, but I can make dev if you prefer. My question would be is once I finish the build issues, but before I add the types, should I make the PR so people can still work coding in JS?

mariana-caldas:
> That makes sense to me, I already changed all the file names though on chore/add-typescript, but I can make dev if you prefer. My question would be is once I finish the build issues, but before I add the types, should I make the PR so people can still work coding in JS? I was just thinking that it would avoid merge conflicts.

More than just preventing merge conflicts, our priority should be to ensure that unfinished work is not pushed to production, Vincent. Here are the recommended steps:

1. Change the current branch `chore/add-typescript` to point to a new branch created from `main`, specifically the `dev` branch.
2. Prioritize resolving any build problems before proceeding with further conversions. This may involve undoing some file renaming if necessary.
3. Once the build problems are solved, you have two options: 
    a. Request a review and merge the `chore/add-typescript` branch into `dev` to initiate a new pull request (PR) from the `dev` branch, such as `chore/add-typescript-phase-2`.
    b. Continue working on the current branch without merging anything into `dev` until the work is completed. This approach allows `main` to remain available for emergency adjustments in simple JavaScript through proper PRs. In case of conflicts between `dev` and `main`, resolving them should not be difficult as we handle everything through PRs. In the worst-case scenario, we can simply copy over any modifications.

Please let me know if this explanation is clear, Vincent!

PR #191, opened 2023-08-22
Fix Who We Are section on mobile

Satoshi-Sh:
| [**189**](https://github.com/MarianaSouza/web-dev-path/issues/189) |

I did.
#### What is this change?
Lowered the threshold to trigger the animation of the developer profile section. 
No.
Checked with the browser mobile simulator. 
After the review.
#### Note:
It would be great if you can check the section on your mobile.

### DIFF
--- a/hooks/useIntersect.js
+++ b/hooks/useIntersect.js
@@ -1,10 +1,14 @@
 import { useState, useEffect, useRef } from 'react';
 
-export const useIntersect = ({ root = null, rootMargin, threshold = 0 }) => {
+export const useIntersect = ({ root = null, rootMargin }) => {
   const [entry, setEntry] = useState({});
   const [node, setNode] = useState(null);
 
   let observer = useRef(null);
+  const threshold = Array.from(
+    { length: 20 },
+    (_, index) => 0.05 + index * 0.05
+  );
 
   useEffect(() => {
     observer = new window.IntersectionObserver(
@@ -23,7 +27,7 @@ export const useIntersect = ({ root = null, rootMargin, threshold = 0 }) => {
         observer.unobserve(node);
       }
     };
-  }, [node, root, rootMargin, threshold]);
+  }, [node, root, rootMargin]);
 
   return [setNode, entry];
 };
--- a/hooks/useIntersect.js
+++ b/hooks/useIntersect.js
@@ -27,7 +27,7 @@ export const useIntersect = ({ root = null, rootMargin }) => {
         observer.unobserve(node);
       }
     };
-  }, [node, root, rootMargin]);
+  }, [node, root, rootMargin, threshold]);
 
   return [setNode, entry];
 };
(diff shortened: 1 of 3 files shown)


mariana-caldas:
Hey, @Satoshi-Sh !

I appreciate the effort in addressing the visibility issue of the `RevealContentContainer` by adjusting the threshold to `0.08`. While this does solve the current problem, I'd like to propose a more flexible solution that can cater to sections of varying heights.

**The Problem:**
The IntersectionObserver's threshold represents the percentage of the target element that is visible within the root's intersection rectangle. A static threshold might not always be optimal as sections grow or shrink.

**Proposed Solution: Multiple Thresholds**
Instead of a single threshold, we can use an array of thresholds, ensuring the callback is triggered at various visibility points.

**Implementation:**

_Hook useIntersect:_

```
import { useState, useEffect, useRef } from 'react';

export const useIntersect = ({ root = null, rootMargin }) => {
  const [entry, setEntry] = useState({});
  const [node, setNode] = useState(null);

  let observer = useRef(null);

  // Create an array of thresholds from 0.01 to 0.99
  const threshold = Array.from({ length: 100 }, (_, index) => (index + 1) * 0.01);

  useEffect(() => {
    observer = new window.IntersectionObserver(
      ([entry]) => {
        setEntry(entry);
      },
      { root, rootMargin, threshold }
    );

    if (node) {
      observer.observe(node);
    }
    return () => {
      observer.disconnect();
      if (node) {
        observer.unobserve(node);
      }
    };
  }, [node, root, rootMargin]);

  return [setNode, entry];
};
```

_Component RevealContentContainer_

The `RevealContentContainer` component receives the hook as an empty object instead of having a hardcoded threshold.

`const [ref, entry] = useIntersect({});`

Please let me know wif that makes sense, Satoshi.

Satoshi-Sh:
Thanks for your help, @mariana-caldas. I changed the code as you instructed. Please have a look when you have time. :+1:

hooks/useIntersect.js
```diff
@@ -23,7 +27,7 @@ export const useIntersect = ({ root = null, rootMargin, threshold = 0 }) => {
         observer.unobserve(node);
       }
     };
-  }, [node, root, rootMargin, threshold]);
+  }, [node, root, rootMargin]);
```
mariana-caldas, line 30: Can you please add back the threshold property here, @Satoshi-Sh ? My example did not have it but it should be included in the dependency array of the `useEffect` hook to ensure that the effect re-runs if the threshold changes. Thanks!

mariana-caldas (changes requested):

mariana-caldas (approved):
LGTM! Thanks, @Satoshi-Sh !

LuSilvaDeveloper (approved):
Everything is working properly here on mobile view.

sergiobtos (approved):
LGTM. For me, it's working. Interesting code. I had never used this way before. Nice.

PR #197, opened 2023-09-11
Allow `BlogPostsContainer` display a dynamic number of cards per row with swipe

mariana-caldas:

Yes

#### What is this change?

* We are adapting the `BlogPostsContainer` component so we can accommodate a dynamic number of cards on the blog page with the swipe functionality.
* We are allowing users to access the 6 latest posts with the swipe functionality.
* The PR template is being updated to include a section to detail how to test a feature or fix it whenever necessary.

No.

1. Check the blog page in prod.  Under the section "Our Success Stories", you see the 4th card is in a second row due to a rule that set the rows would be formed by 3 posts.

Screen Shot 2023-09-11 at 4 04 39 PM

2. When running this branch locally, that's the expected result on the desktop. Notice the swipe functionality indicates, even on the desktop, that there are more posts on that row.

Screen Shot 2023-09-11 at 4 05 16 PM

Check the point above.

After review


Satoshi-Sh (approved):
Great catch, Mariana.  Now we can add more blog articles and they get displayed nicely.

Thanks for adding the test section to the pull request template, which makes reviewing easier.

LuSilvaDeveloper:
It looks perfect. The blog page looks consistent with all the sections having the same behaviour.  The swipe functionality makes it look more clean.

LuSilvaDeveloper (approved):

components/blog/BlogPostsContainer/index.js
```diff
@@ -41,10 +41,9 @@ const BlogPostsContainer = ({
           </Container>
         ) : null}
         {
-          // put in rows of 3 if more than 3 posts (for swipable cards)
           swipe ? (
             <>
-              {[...splitPosts(posts, 3)].map((p, index) => (
+              {[...splitPosts(posts, posts.length)].map((p, index) => (
```
cherylli, line 46: I think `splitPosts` is not required now it's always "splits" into a length if itself (`posts.length`) which means no splitting was actually done. It seems 

```js
{[posts].map((p, index) => (
 <BlogCardsColumns key={index} cards={p} />
))}
```
works the same

pages/blog/index.js
cherylli, line 1: it's from previous PR but line 6: `// import styles from '@/styles/Blog.module.scss';` should probably be removed

cherylli (approved):
works fine but I left a couple of comments

components/blog/BlogPostsContainer/index.js
mariana-caldas: Good catch, @cherylli ! Thanks!

PR #210, opened 2024-01-16
Feature/add theming

LuSilvaDeveloper:
| [**165**](https://github.com/MarianaSouza/web-dev-path/issues/165) |

Yes.

#### What is this change?

Adding theming to the project to have centralized all SASS variables previously spread across _mixins.js and _variables.js into themeConfig.js for better organization. 

Some questions:

- We have some components that were using both SASS and normal CSS, for example, inside `styles.js` from `Hero` we have the `HeaderContent` which was previously using SASS for `color` and normal CSS for `padding-top`. The best approach should be keeping the of `padding-top` inside the `styles.js` file, and `color` we keep it inside `themeConfig.js`?

Screenshot 2024-01-16 at 6 40 12 AM

- Should we pass all attributes inside `globals.scss` to `themeConfig.js` on `GlobalStyles` or should we keep only the attributes that were using SASS inside `themeConfig.js`, similarly to the previous question?

Screenshot 2024-01-16 at 6 50 00 AM

Screenshot 2024-01-16 at 6 50 20 AM

The next step would be passing these attributes to all components, making sure everything is working.

Mostly misunderstanding on my part regarding the issue

Run on different browsers.

NA


themeConfig.js
```diff
@@ -0,0 +1,92 @@
+import { createGlobalStyle } from 'styled-components';
+
+export const lightTheme = {
+  breakpoints: {
```
cherylli, line 4: wouldn't the light and dark theme have the same breakpoints?

styles/globals.scss
```diff
@@ -5,44 +5,44 @@ html,
 body {
   padding: 0;
   margin: 0;
-  font-family: $copy-font, sans-serif;
-  color: $primary-content-color;
+  // font-family: $copy-font, sans-serif;
```
cherylli, line 8: can these commented code be removed? or are we keeping them for something?

components/layout/Footer/styles.js
```diff
@@ -14,7 +15,7 @@ const Inner = styled(Container)`
 
   //media query mixins
   ${m.tablet(css`
-    display: flex;
+    display: ${({ theme }) => theme.display};
```
cherylli, line 18: does changing theme change the layout too? I have an impression that it would just be the colour

components/layout/Hero/styles.js
```diff
@@ -17,14 +17,14 @@ const Header = styled.div`
 `;
 const HeaderContent = styled(Container)`
   z-index: 10;
-  color: ${$white};
+  color: ${({ theme }) => theme.colors.white};
```
cherylli, line 20: would this always be white regardless of the theme? or is "white" a different shade of white depending on the theme?

cherylli (commented):
Thanks. It works locally without any errors but I just have a few questions. Please see inline comments 

Also I see a lot of commented code, are they all needed?

styles/globals.scss
LuSilvaDeveloper: Sure. I'm removing it.

components/layout/Hero/styles.js
LuSilvaDev: It would be always white. But the point here was removing the SASS and put this color inside themeConfig

themeConfig.js
LuSilvaDeveloper: I think so. But for the first moment, it's supposed to get the variables using SASS and put it on the themeConfig, and I'm just leaving the darkTheme ready for further implementation

components/layout/Hero/styles.js
LuSilvaDeveloper: Yes. But The point is changing from SASS to have this variable inside themeConfig.

cherylli: I think it's fine for now. This allows us to have slightly different shade of white in different themes. Also, since we don't really have another theme yet we don't really know what it would be in another theme, it could be like a secondary color

components/layout/Footer/styles.js
```diff
@@ -76,25 +81,27 @@ const SocialList = styled.ul`
...
+    @media (min-width: ${props.theme.breakpoints.tablet}) {
+      justify-content: end;
+    }
+  `}
 `;
 
 const SocialItem = styled.li`
   cursor: pointer;
-  ${m.transition('all 0.3s ease')};
+  transition: all 0.3s ease; // is it the same as the transition mixin?
```
cherylli, line 93: I don't really know mixin but looks like you'll have to use `@include`
https://stackoverflow.com/questions/62999437/how-to-create-a-transform-mixin-in-scss-that-uses-several-arguments

someone correct me if I'm wrong

themeConfig.js
```diff
@@ -0,0 +1,92 @@
...
+  body: {
+    margin: '0',
+    padding: '0',
+  },
+  lists: {
+    ul: {
+      listStyle: 'none'
+    }
+  },
+  a: { // change the name
```
cherylli, line 44: whats "change the name"?

cherylli (commented):
everything works fine locally, I just have a few questions

themeConfig.js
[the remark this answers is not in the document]
LuSilvaDeveloper: I think it depends on the design that we want for the app. It should work fine since it makes everything global on the application. Do you think is it better to put it in the styles folder?

themeConfig.js
LuSilvaDeveloper: I just forgot to delete this comment

themeConfig.js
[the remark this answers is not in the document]
cherylli: I think we should put in the styles folder as it's style related. I don't think you need to put it in the root for it to be globally accessible

mariana-caldas (changes requested):
We are getting close, @LuSilvaDeveloper ! Great job!

Besides our comments, it's important that at the end, we have no `.scss` files inside our `styles` folder since any global style should be implemented within the `themeConfig.js` file.

I answered the questions you added to the PR body on Slack. Can you please remove them from the PR body? Ideally, the PR description will give us information in the future so, I'd say ask things on Slack or use the comments.

Can you please request our review again after you implement the changes? Thanks, so much, Lucas!

components/containers/TwoColumn/styles.js
```diff
@@ -106,10 +104,11 @@ const AboutUs = {
...
   height: 10rem;
 
-  //media query mixins
-  ${m.desktop(css`
-    height: 12rem;
-  `)}
+  ${props => css`
+    @media (min-width: ${props.theme.breakpoints.desktop}) {
+      eight: 12rem;
+    }
```
mariana-caldas, line 110: Please fix the typo "height" here and test it out with different breakpoints.

components/layout/Footer/styles.js
```diff
@@ -76,25 +81,27 @@ const SocialList = styled.ul`
...
   }
 
-  //media query mixins
-  ${m.tablet(css`
-    padding-left: 28px;
-    max-width: 100%;
-  `)}
+  ${props => css`
+    @media (min-width: ${props.theme.breakpoints.tablet}) {
+      padding-left: 28px;
```
mariana-caldas, line 101: Let's keep using only rem not px, please.

components/layout/Footer/styles.js
mariana-caldas: I think since we're using styled-components, the implementation is a little bit different, @cherylli.

We should have the variables and mixins inside themeConfig as well, @LuSilvaDeveloper, so we can get rid of any scss file. Then, we indeed use the transition from the mixin function we get from the themeConfig file.

themeConfig.js
```diff
@@ -0,0 +1,92 @@
...
+  text: '#363537',
+  background: '#023047',
+  fontStyle: {
+    italic: 'italic',
+  },
+}
+
+// The darkTheme is a theme that we will have on the future
+// for now it's empty
+export const darkTheme = {};
```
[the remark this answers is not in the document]
LuSilvaDeveloper: It makes a lot of sense! thank you!

components/containers/TwoColumn/styles.js
LuSilvaDeveloper: Done!

LuSilvaDeveloper:
I think I've done the changes as required. Please let me know if it's correct. Also, once is reviewed, I will delete the globals.scss file.

cherylli:
thanks! Everything looks good to me!

cherylli (approved):
thanks! Everything looks good to me!

components/buttons/SubmitButton/index.js
```diff
@@ -1,6 +1,6 @@
 import S from './styles';
 
-export function SubmitButton({ label, disabled }) {
+export function SubmitButton({ label, disabled }) { {}
```
mariana-caldas, line 3: Can you help us to confirm why we need this {} on line 3, @LuSilvaDeveloper ?

mariana-caldas (approved):
Such a great job, @LuSilvaDeveloper ! 🚀 
I will leave this PR approved, however, before merging it. Can you please take a look at the comment on SubmitButton/index? Thanks so much!

components/buttons/SubmitButton/index.js
LuSilvaDeveloper: Actually, we don't need it. It's unnecessary.

Issue #238, opened 2025-05-12
Discussion: Migrate Away from styled-components Due to Package Entering Maintenance Mode


cherylli:
**What do we need to build or fix?**
We need to discuss and plan the migration away from styled-components, which is now in [maintenance mode](https://opencollective.com/styled-components/updates/thank-you).
Although there's no immediate need to migrate away from it, continuing to rely on it poses long-term risks in terms of support, compatibility with new React/Next.js features, and community adoption.

**Technical details**
The goal is to identify and adopt a styling solution that:

- Works seamlessly with our current tech stack (e.g. Next.js, App Router)
- Provides better developer experience (DX), performance, and maintainability

**Approach suggestions**
Please research alternative styling solutions that are actively maintained and align with our tech stack and long-term goals.

Bring your findings to this discussion, including pros and cons of options like Tailwind CSS, CSS Modules, Panda CSS, Vanilla Extract, etc.

We’ll use this issue to weigh trade-offs and agree on a direction before starting any migration work.

**Acceptance criteria**

This is a discussion issue

cherylli:
@Satoshi-Sh @LuSilvaDeveloper @Web-Dev-Path/project-managers-and-scrum-masters

cherylli:
Safest option is probably back to CSS Modules with sass (what we had before), but we need to figure out theming with it (or do we need theming? since we really only have 1 theme)

Satoshi-Sh:
I personally enjoy using Tailwind.css, but it makes more sense to go back to the CSS Modules with Sass here. Maintaining a project is quite a task. 😅

oluwatobiss:
Here are some advantages Tailwind and CSS Modules have over one another:

### Tailwind 

- **Faster development:** Tailwind allows style composition on the go (while developing the UIs).
- **Faster UI maintenance:** Having styles directly on each UI makes understanding and maintaining each element’s style faster than switching between files to find the styles.
- **No names to invent:** Tailwind allows you to focus more on creating beautiful designs rather than considering the name to assign to an element.
- **Predefined designs:** Tailwind’s inbuilt design features make moving from concept to implementation faster.

### CSS Module

- **Future proof:** There’s that essential confidence that we are using a tool that will not suddenly go into maintenance mode.
- **Reusability:** CSS modules allow using the same style in multiple modules or for various elements. (The Tailwind workaround for this can be messy and strenuous to maintain).
[...]

PR #239, opened 2025-06-17
Migrating project back to CSS Modules with SCSS from style-components

cherylli:

Yes

#### What is this change?
The main purpose of this change is to test out if css modules and styled-components can co-exist, and migrate fully to CSS modules in stages.

In this PR,
- `ContactUsCards` is changed to using CSS modules
- brought back the old _mixins, and _variables.scss 
- removed color variables in _variables.scss and move to using CSS variables for theming (`styles/themes.scss`)
- variable $white/--white, was renamed to something more descriptive and compatible with theming 

CSS variables are used for theming mainly due to the fact that it can be changed during runtime, unlike scss varibles, theme is set during compile time.

To change theme we would set the data-theme attribute on the document like so
`document.documentElement.setAttribute('data-theme', 'dark');`
for non default themes

I also want to use a different file structure — specifically, placing all style-related files close to their respective components, as opposed to keeping all .scss files in a single styles folder, which becomes hard to manage as the project grows.

This is a link of our repo just before conversion to styled-components
https://github.dev/Web-Dev-Path/web-dev-path/blob/e5fd0ab9b45bc32f2564c81d635d80ca4d9241c3/styles/ContactUsCards.module.scss

No

Dark theme is populated with placeholder colors, you can test it out by adding a useEffect in any page/component to set the data-theme attribute

After review

### DIFF
--- a/pages/contact.js
+++ b/pages/contact.js
@@ -1,6 +1,6 @@
 import { useState } from 'react';
 import ContactUsFormSubscribe from '@/components/ContactUs';
-import Index from '@/components/ContactUs/ContactUsCards';
+import ContactUsCards from '@/components/ContactUs/ContactUsCards';
 import S from '@/styles/pages/contactStyles';
 
 export default function ContactUs() {
@@ -23,7 +23,7 @@ export default function ContactUs() {
           </S.ResponseMessage>
         </S.FormAndDecorations>
       </S.ContactUsContainer>
-      <Index />
+      <ContactUsCards />
     </>
   );
 }


pages/contact.js
```diff
@@ -23,7 +23,7 @@ export default function ContactUs() {
           </S.ResponseMessage>
         </S.FormAndDecorations>
       </S.ContactUsContainer>
-      <ContactUsCards />
+      <Index />
```
oluwatobiss, line 26: Any reason why the `Index` naming choice is preferred here? `ContactUsCards` (or `ProjectInfoCards`) looks more descriptive.

oluwatobiss (commented):
Great work here, @cherylli! I'm curious about the `Index` naming choice. Please see the inline comment.

pages/contact.js
cherylli: no, it was the auto renaming from my IDE when I rename the file, I thought I fixed it. Thanks for noticing

oluwatobiss (approved):
LGTM

Satoshi-Sh (approved):
Looks good. 

Is it better to have a list to track which part we completed the migration on #238 or somewhere? 

## Blog 

- [ ] AuthorBio
- [ ] BlogPostContainer

## ContactUs
- [x]  ContactUsCards
- [ ]  ContactUsForm

briangesteban (approved):
This looks good already. Just a suggestion tho, maybe we could modularize the partials, like instead of having all variables in _variables.scss we could put it as _colors.scss, _breakpoints.scss, _typgraphy.scss, etc. Same with mixins, it could be like _media-queries.scss, etc. and also adding fallback fonts. Other than that everything looks great to me.

cherylli:
> This looks good already. Just a suggestion tho, maybe we could modularize the partials, like instead of having all variables in _variables.scss we could put it as _colors.scss, _breakpoints.scss, _typgraphy.scss, etc. Same with mixins, it could be like _media-queries.scss, etc. and also adding fallback fonts. Other than that everything looks great to me.

I'm not sure if putting colors in scss will work for theming since it's compiled at build time, thats why I have them as normal css variables, but feel free to test it and let us know

briangesteban:
> > This looks good already. Just a suggestion tho, maybe we could modularize the partials, like instead of having all variables in _variables.scss we could put it as _colors.scss, _breakpoints.scss, _typgraphy.scss, etc. Same with mixins, it could be like _media-queries.scss, etc. and also adding fallback fonts. Other than that everything looks great to me.
> 
> I'm not sure if putting colors in scss will work for theming since it's compiled at build time, thats why I have them as normal css variables, but feel free to test it and let us know

Yes, you are absolutely right but there's a work around for it that we could do. We could implement a CSS/SCSS variables hybrid. Theming will still use the CSS variables, then we will use interpolated SCSS variables as a value for those CSS variables. 

SASS have an interpolation method, which looks like this `#{}`. Just wrap the anything(like a variable) inside the curly braces `#{$scss-color-var}`. Here's an example code snippet.

**Colors Partial File**
```
$primary-color: #ffffff;
$accent-color #000000;

$primary-color--dark: #000000;
$accent-color--dark: #ffffff;
```

**Theme Partial File**
```
@use "./colors.scss" as *;

:root {
--primary-surface-color:  #{$primary-color}
--primary-text-color: #{$accent-color}
}

.dark-theme {
--primary-surface-color:  #{$primary-color--dark}
--primary-text-color: #{$accent-color--dark}
}
```

**Reference**
[SASS Interpolation Method (to quoteless string)](https://sass-lang.com/documentation/interpolation/)

PR #242, opened 2025-06-19
CSS Modules Migration for ButtonLink Component

Satoshi-Sh:

This PR depends on #239
Yes
#### What is this change?
CSS Modules Migration for ButtonLink Component

You could check the styling behavior of Contact Us button on `/about` page.

Screenshot from 2025-06-19 08-04-07

After review and once PR #239 is merged.


cherylli:
@Satoshi-Sh could you please pull the latest main and push to your branch?

components/buttons/ButtonLink/index.js
```diff
@@ -1,21 +1,23 @@
...
-/*  The component supports the use of target with the property "openNewTab" to open the
- link in a new tab.*/
 export default function ButtonLink({
   link,
   children,
-  $colorScheme,
+  customClassName,
   openNewTab,
 }) {
+  let buttonClass = customClassName
```
cherylli, line 9: we should probably use `const` here or just use it inline

Satoshi-Sh:
Thanks for reviewing this in detail. @cherylli

briangesteban (changes requested):
**Fix Request**
- Need to fix buttons on the Homepage as well. The background color and transition is not working properly.

**Improvement Suggestion**
```
  const buttonClass = customBtnClass
    ? `${styles.buttonLink} ${styles[customBtnClass]}`
    : styles.buttonLink;
```
This code snippet above seems to be re-usable, and might be use in the SubmitButton component. We could create a helper function under utils for this, something like getButtonStyle or something similar.

briangesteban (approved):
Looks good to me now. Everything's working back smoothly.

PR #248, opened 2025-07-08
Refactor/to-scss-blog-post-container

briangesteban:

Yes

#### What is this change?
- Remove theming from styled-components in BlogPostContainer.

None

N/A

N/A

After the reviews.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -163,5 +163,5 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   - ButtonLink
   - SubmitButton
   - AuthorBio
-  
+  - BlogPostContainer
 - Updated ContactUsForm's checkbox wrapper from div to label to enhance its accessibility
--- a/components/blog/BlogPostContainer/index.js
+++ b/components/blog/BlogPostContainer/index.js
@@ -3,10 +3,8 @@ import TwoColumn from '@/components/containers/TwoColumn';
 import AuthorBio from '@/components/blog/AuthorBio';
 import RelatedOrLatestPosts from '@/components/blog/RelatedPosts';
 import PostContent from '@/components/blog/PostContent';
-import { useTheme } from 'styled-components';
 
 const BlogPostContainer = ({ post, relatedPosts, latestPosts }) => {
-  const theme = useTheme();
   const { user } = post;
   return (
     <>
@@ -28,8 +26,8 @@ const BlogPostContainer = ({ post, relatedPosts, latestPosts }) => {
           content='Feel free to contact us.'
           image='/images/svg/square-brackets.svg'
           altTag='Hashtag'
-          color={theme.colors.primaryContent}
-          bgColor={theme.colors.white}
+          color='#292929' // Temporarily set static values, to prevent breaking TwoColumn components.
+          bgColor='#ffffff' // We might remove these two props when we implement our own ThemeProvider.
           $contentType='questions'
           link='/contact'
           linkText='Contact us'


components/blog/BlogPostContainer/index.js
```diff
@@ -28,8 +26,8 @@ const BlogPostContainer = ({ post, relatedPosts, latestPosts }) => {
           content='Feel free to contact us.'
           image='/images/svg/square-brackets.svg'
           altTag='Hashtag'
-          color={theme.colors.primaryContent}
-          bgColor={theme.colors.white}
+          color='#292929' // Temporarily set static values, to prevent breaking TwoColumn components.
```
cherylli, line 29: we should use variables already in `styles/themes.scss` instead of hardcoding them
If the variables don't already exist we can add them (I didn't add all the old variables back, so we can add as we go instead of leaving all unused values in the file)

cherylli: I got it to work -
1. create a scss file for BlogPostContainer
2. export the sass variable from it 
```
@use '@/styles/variables' as *;

:export {
  bgColor: $bgColor;
}
```
3. Import 
`import {bgColor} from './BlogPostContainer.module.scss'`

4. use it
`bgColor= {bgColor}`

test with `document.documentElement.setAttribute('data-theme', 'dark');`

It does actually change to dark theme bgColor


Unrelated to this change

This should be Square Brackets `altTag='Hashtag'`

mariana-caldas: What if we change the PR scope and adjust the TwoColumns components as well?

cherylli: yeah we can, we still need to solve the same "issue" though

briangesteban:
> how can theming get rid of a prop to pass in a custom color, which could be different from either (or any) theme values?
> 
> Although for this particular component, we could just remove bgColor and $color props are they are same as the default for TwoColumn

For the theming(ThemeProvider) to work, we should have color stylings live in the stylesheet, by using the theme tokens (CSS Variables) applied to component/elements styles.

For passing custom color/style issue, we could instead create a helper function or enums to have restricted options of class names to be pass down to that component. We could create many custom classes as we want since it will all live on the stylesheets and won't be on our way on the component logic. Yes, we're still using props to be able to give it a custom color/style, but we've lessen it to one prop and only using html/js logic by creating a helper function that returns a pre-defined classNames. All stylings properties/values are still inside the stylesheet, which won't break the theming.

Here's an example of what we could possibly do:

**Helper function to return a defined className for TwoColumns:**
```
// Could be a Switch case if there's a lot to define
const getStylesTwoColumns = (style) => {
if(style === "tc-red") return ".two-columns--red";
else if(style === "tc-blue" return ".two-columns--blue";
else return ".tc-default";
}
```
**Style sheet for TwoColumns:**
```
@use "./colors" as *;

.twoColumns {
someProps: someValues;
}

// Defined custom stylings
.tc-red {
 background-color: var(--color-red);
}

.tc-blue {
background-color: var(--color-blue);
}

.tc-default {
background-color var(--color-white);
border-radius: 2rem;
width: 50%;
}
```

**Assuming TwoColumns component have something like this:**
```
const TowColumns = ({someProp, anotherProp, customStyle = "tc-default", children}) => {
return (
 <section className={[styles.twoColumns, customStyle ]}>
    <h2>{someProp}</h2>
    <p>{anotherProp}</p>
    {children}
 </section>
)
}
```

**Implementation of the helper function to assign the custom class:**
```
<TwoColumns someProp={heading} anotherProp={content} style={getStyles("tc-blue")}
  {children}
</TwoColumns>
```

cherylli:
I suggest for this component, we remove the props and use the default, as the props are just the defaults.

We should proceed on the migration task same as the other components, using css variables. 

We haven't actually have a plan to include another theme. And honestly I don't see having themes is normal for a "company/organisation website" in oppose to a website people uses like github/jira. 

We are complicating this task by trying to fit into an unplanned feature. If we do ever implement theming, I suggest we can look at the Theme section in the CSS Modules repo. https://github.com/css-modules/css-modules/blob/master/docs/theming.md

This looks a lot more straightforward. 

What do you think @mariana-caldas

Satoshi-Sh (approved):
Looks good to me, and thanks for the reference for altTag.

```
image='/images/svg/square-brackets.svg' // Decorative Image
altTag='' // Intentionally left empty to comply with W3C accessibility standards.
```
You could delete the comments on line 27/28

You can also update other TwoColumn's altTag in this PR.

cherylli:
> Looks good to me, and thanks for the reference for altTag.
> 
> ```
> image='/images/svg/square-brackets.svg' // Decorative Image
> altTag='' // Intentionally left empty to comply with W3C accessibility standards.
> ```
> 
> You could delete the comments on line 27/28
> 
> You can also update other TwoColumn's altTag in this PR.

I agree these comments can be removed, and there are still conflicts need to be resolved

briangesteban:
I've resolved the conflicts, removed the comments, and updated the other altTags from TwoColumn components to comply with the W3C standards.

If I've missed something, please let me know.

Thank you @cherylli and @Satoshi-Sh for the review!

CHANGELOG.md
```diff
@@ -160,7 +160,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 - Migrating styles from Styled Components to CSS Modules
-  - ContactUsCard
```
Satoshi-Sh, line 163: You are right that we haven't finished `ContactUsCard`. This should be `ContactUsCards`

Satoshi-Sh (approved):
Thanks for updating the alt tags. Looks good to me.

PR #249, opened 2025-07-09
Refactor/to scss blog posts container

briangesteban:

Yes

#### What is this change?
Migrate BlogPostsContainer from styled-components to SCSS module.

None

N/A

N/A

After the reviews.

#### Image Reference
ref-blogs
ref-blogs-all
ref-blogs-backlink

### DIFF
--- a/components/blog/BlogPostsContainer/BlogPostsContainer.module.scss
+++ b/components/blog/BlogPostsContainer/BlogPostsContainer.module.scss
@@ -26,20 +26,6 @@
   @include large-desktop {
     flex-basis: 32%;
   }
-
-  // I'm not sure why we have this, but it seems unecessary since it doesn't affect the layout.
-  // It might be a legacy feature or I might be wrong.
-
-  // &::after {
-  //   @include tablet {
-  //     content: '';
-  //     flex: auto;
-  //   }
-
-  //   @include desktop {
-  //     content: unset;
-  //   }
-  // }
 }
 
 .bottomLink {


briangesteban:
Also, just a heads up. I've noticed something unusual while working with this PR. I compared this PR with the live prod to make sure it's still styled and working accordingly. But there seems to be some broken links on the live prod, that is not broken on the staging(PR).

**To replicate the issue**
From the live prod: navigate to blogs page > under Latest Posts > Why AI struggles with "no" > click the Read more link
> It leads to a 404 page

From the staging: navigate to blogs page > under Latest Posts > Why AI struggles with "no" > click the Read more link
> It leads to the actual blog

I can't find the cause, but it's quite unusual that the live prod leads to 404 while the staging leads to the actual blog. And it's only affecting some links, not all of it.

Satoshi-Sh:
> there seems to be some broken links on the live prod, that is not broken on the staging(PR).

It seems that the production hit the DevTo API rate limit when generating individual blog pages. I don't know why staiging is not affected.

cherylli:
In prod, those pages are SSG, while in staging (and dev mode), they are dynamic
image


I thought we fixed this issue, maybe dev.to reduced the limit again. Might have to find some workaround for the API rate limit, or just make those pages dynamic

Satoshi-Sh (commented):
I played with tha part you commeneted out and didn't see any effects. I believe you can delete it. 

```
// I'm not sure why we have this, but it seems unecessary since it doesn't affect the layout.
  // It might be a legacy feature or I might be wrong.

  // &::after {
  //   @include tablet {
  //     content: '';
  //     flex: auto;
  //   }

  //   @include desktop {
  //     content: unset;
  //   }
  // }
}
```

~One small thing is about the class naming.  Hypenated(`blog-container`) is better than camel cases(`blogContainer`) as~  ~it is scss.  My first PR was using camel cases though.~ Never mind this one. I found camel cases are recommended here. 

Screenshot from 2025-07-09 15-36-51

Feel free to assign yourself to a PR, request reviewers (we use web-devpath/development team), and add related labels (migration-task).

cherylli:
> I'm not familiar with DevTo's API, but is it possible for us to fetch all the blogs one time and save it on a JSON file in our repo; then we will just automate the API call to fetch new data to update the JSON file when it detects changes (added, deleted, updated)? That way we will not do an API call every time a user navigates to our blog page or when we develop/test our web app.

we are using ISR atm, it's supposed to cache everything in build time and revalidate automatically every day (when someone access it through the link) , I guess the revalidate part might not be working. Because I don't have access to the deployment I can't really look at the logs etc to find the exact cause

cherylli:
I tested locally using build, then serve from the built file. All posts was working, so the code is fine, and probably not API rate limit (unless they have different rate limit at different times).

@mariana-caldas was there any error on netlify for the last build? or any runtime error?

Satoshi-Sh (approved):
Looks good to me. Thanks for your work, Brian :+1:

cherylli (approved):

PR #251, opened 2025-07-09
Refactor/to-scss-reveal-content-container

briangesteban:

Yes

#### What is this change?
- Migrate `RevealContentContainer` component from styled-components to SCSS module.

None

N/A

Reload browser or scroll down to check first load transitions.

After the reviews.

#### Image Reference
Not available since it's non-static update. It's the transition from first load.


components/containers/RevealContentContainer/index.js
```diff
@@ -1,11 +1,12 @@
 import { useEffect, useState } from 'react';
 import { useIntersect } from '@/hooks/useIntersect';
-import S from './styles';
+import styles from './RevealContentContainer.module.scss';
 
 const RevealContentContainer = ({ children }) => {
   const [ref, entry] = useIntersect({});
   const [firstLoad, setFirstLoad] = useState(true);
   const [hiddenStyle, setHiddenStyle] = useState(true);
+  const hiddenStyleClass = hiddenStyle ? styles.sectionHidden : null; // return sectionHidden class name if hiddenStyle is true.
```
cherylli, line 9: We can remove this comment. It's not necessary to add comments to self-explanatory code as people reading the code are usually also developers

Another issue with this line is if it's set to null, it will has class="null" in the html like so

Image

briangesteban: Hi @cherylli ! Thanks for the review. I've fix the null set on class name, and remove the unnecessary comment.

shayla-develops-webs (commented):
Looks great!

oluwatobiss (approved):
`RevealContentContainer` looks good. Thanks!

PR #257, opened 2025-07-16
Refactor/to scss member

Satoshi-Sh:

Yes
#### What is this change?
CSS modules migration for Member component

Please check the member part stylings on About page.

Screenshot from 2025-07-15 19-59-04

After review


components/containers/Member/Member.module.scss
```diff
@@ -0,0 +1,99 @@
...
+    min-width: 20%;
+    max-width: 20%;
+    margin: 1.5rem;
+  }
+}
+
+.Name {
+  font-style: italic;
+  font-family: var(--font-heading);
+  font-size: 1.5rem !important;
```
cherylli, line 29: it looks like this !important can be removed

components/containers/Member/Member.module.scss
```diff
@@ -0,0 +1,99 @@
...
+  text-align: left;
+  line-height: normal;
+  margin: 0px !important;
+
+  @include tablet {
+    font-size: 2rem;
+  }
+}
+
+.H3 {
```
cherylli, line 48: is the `H3` class used? If not we can remove it

components/containers/Member/Member.module.scss
```diff
@@ -0,0 +1,99 @@
...
+  font-family: var(--font-heading);
+  font-size: 1.5rem !important;
+  color: var(--color-primary-content);
+  margin-bottom: 0px;
+  line-height: unset;
+}
+
+%headingStyle {
+  font-family: var(--font-heading);
+  font-size: 1rem !important;
```
cherylli, line 37: I think this !important overrides, the 2rem in `@include tablet` below, so we can remove "@include tablet" and remove !important as well

cherylli (changes requested):
I didn't comment on all the !important but I think this file as way too many !important. I know it's copied from the styles but maybe we should see if we can remove most if not all of the !important

shayla-develops-webs (approved):
Looks good Satoshi!

oluwatobiss (commented):
Why don’t we stick with camelCasing as @cherylli recommended on Slack?

oluwatobiss (approved):
All looks great! Thanks @Satoshi-Sh.

PR #258, opened 2025-07-19
Refactor/to scss container

Satoshi-Sh:

Yes
#### What is this change?
- CSS modeeuls migration for Container
- Additionally, I updated Member as I'd noticed the $heading-font was defined in `_variables.scss`

Please check the place Container used 

Screenshot from 2025-07-19 08-15-50

After review

### DIFF
--- a/components/containers/Container/index.js
+++ b/components/containers/Container/index.js
@@ -1,9 +1,9 @@
-import S from './styles';
+import styles from './Container.module.scss';
 
-export default function Container({ className, children, styles }) {
+export default function Container({ className, children, _styles }) {
   return (
-    <S.Container className={className} style={styles}>
+    <div className={`${styles.container} ${className || ''}`} style={_styles}>
       {children}
-    </S.Container>
+    </div>
   );
 }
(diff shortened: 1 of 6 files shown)


components/containers/Container/index.js
```diff
@@ -1,9 +1,9 @@
-import S from './styles';
+import styles from './Container.module.scss';
 
-export default function Container({ className, children, styles }) {
+export default function Container({ className, children, _styles }) {
```
shayla-develops-webs, line 3: I would probably just use the React standard naming convention for inline styles and keep just 'style' over '_styles' to avoid confusion to other developers who expect inline styles to be passed via style.

components/containers/Container/index.js
```diff
@@ -1,9 +1,9 @@
...
   return (
-    <S.Container className={className} style={styles}>
+    <div className={`${styles.container} ${className || ''}`} style={_styles}>
```
shayla-develops-webs, line 5: Same thing here. I would use style={style}

shayla-develops-webs (approved):
Approving this as-is since it works perfectly and isn’t a blocking issue.
Just a heads-up for next time: using the prop name style (without underscore) would align better with React’s standard conventions for inline styles and improve clarity for future maintainers.
Thanks for the solid work on this!

cherylli (approved):
thanks!

PR #263, opened 2025-07-22
Fix "Row-Reverse" Styling Bug

Satoshi-Sh:
| [**262**](https://github.com/MarianaSouza/web-dev-path/issues/262) |

Yes
#### What is this change?
The styling bug was introduced in my last PR #258 . Reverting the Container prop name(styles) fixed the issue.

Please check the styling both home and about page. Two columns with `row-reverse` .

Screenshot from 2025-07-22 06-54-52

 
After review


cherylli:
I suggest converting this to CSS modules while we are at it to make sure the fix will work after the conversion. Thoughts?

cherylli (approved):
Thanks for the fix

components/containers/Container/index.js
```diff
@@ -1,8 +1,8 @@
-import styles from './Container.module.scss';
```
briangesteban, line 1: Just a suggestion, to maintain consistency on naming our imports of SCSS modules, we could maintain this named import as is and rename the prop name instead as we destructure the prop object. This way we could avoid the name conflict between the import and the props, and also make the two distinguishable.  

`export default function Container({ className, children, styles: propStyles })`

briangesteban (approved):
Looks good, bug is fixed. Thanks @Satoshi-Sh !

components/containers/Container/index.js
Satoshi-Sh: Good idea. I will use this

oluwatobiss (approved):
LGTM

PR #264, opened 2025-07-23
Fix styles barrel import, extract :root to globals.scss

cherylli:
yes

#### What is this change?

Fix styles barrel import
extract :root to globals.scss

:root should only be imported once in _app, so it should be taken off barrel import 

no

n/a


Satoshi-Sh:
Thanks. It's really handy.

`BlogPostsContainer.modules.scss` needs to be updated in this PR.

styles/themes.scss
oluwatobiss, line 1: Shouldn’t the `[data-theme='dark']` rule (and the comment) also move into `globals.css`? As is, this file is unused. Consider removing it if `[data-theme='dark']` moves.

oluwatobiss: Here’s one perspective on the use of the global and theme stylesheets:

- **Global:** Used for global rules that affect all elements by default. For instance, a CSS Reset for the entire app.
- **Theme:** Used for defining brand-specific opt-in styles that can be reused across multiple elements. For example, link colors, font types, or styles for dark and light modes.

Following this perspective, the declarations for `:root` and `[data-theme='dark']` should remain in the `themes.scsss` file. While `globals.scss` should be used for any global ruleset (if we have any).

Or what’s your view on the purposes of the two stylesheets?

oluwatobiss (approved):
Duplicate `:root` imports fixed. Thanks!

PR #273, opened 2025-08-09
Refactor/to scss layout

Satoshi-Sh:

Yes

#### What is this change?
CSS modules migration for Layout component

Please check the main styling on any page.

After reviews


cherylli (approved):

oluwatobiss:
> We could delete the class, but how can we make sure the rules not affecting anything?

It has no impact on all the pages I’ve checked. However, even if it does, it seems odd to change the stacking context of the website’s main container. It can make debugging challenging. This case of uncertainty about whether the rule has any side effects is a good example of the unintended consequences that such implementation can cause. Since we are working on the style conversion, it may be best to delete it now so we can determine its impact as we progress.

PR #276, opened 2025-08-17
Refactor/to scss nav

Satoshi-Sh:

Yes
#### What is this change?
- CSS moduels migration for Nav
- Fixed active state bug on navigation. When the window width pass the desktop breaking point, the active state should be turned to false.
Screenshot from 2025-08-16 21-32-35

On prod or main branch, press the hamburger menu on the narrow window and then make the window wider.  

Please check the navigation styling behavior.

After reviews


components/layout/Nav/index.js
```diff
@@ -3,14 +3,15 @@ import { useRouter } from 'next/router';
...
-import S from './styles';
+import styles from './Nav.module.scss';
 
 export default function Nav() {
   const router = useRouter();
   const [active, setActive] = useState(false);
   const [isSticky, setIsSticky] = useState(false);
   const headerRef = useRef();
   const containerRef = useRef();
+  const DESKTOP_BREAKPOINT = 1024;
```
cherylli, line 14: is it possible to use the desktop-breakpoint already defined in `styles/_variables`?

Satoshi-Sh: I just tried several approaches to use the breakpoint from `_variables.scss`, but none worked and it feels like overengineering.  

Let me know if you have a solution.

PR #303, opened 2026-04-09
Fix/security input validation

shayla-develops-webs:
| [**250**](https://github.com/MarianaSouza/web-dev-path/issues/250) |

Yes

#### What is this change?

Fixed security vulnerabilities related to user input handling across three files:

- Replaced dangerouslySetInnerHTML with plain text rendering in NewsletterForm to fix an XSS vulnerability
 - Added client-side email format validation to NewsletterForm
- Escaped all user inputs in sendEmail.js using encode() from html-entities to prevent HTML injection
- Added server-side email format, name length, and message length validation in validateReCaptcha.js

During local setup, the dev server would not start due to a TypeError: withPWA is not a function error in next.config.js. This was caused by a breaking change in the next-pwa package API. I fixed the import syntax to match the installed version before proceeding with the changes. No new dependencies were required for the actual security fixes.

On the current live site, go to the newsletter form and submit <script>alert('xss')</script> in the email field. It will pass through without any format validation
On the contact form, submit `">` in the name or message field. The input gets inserted directly into the outgoing email HTML template with no escaping, allowing injected HTML to render in the email client.

On the newsletter form, submit <script>alert('xss')</script> in the email field. It should be rejected with an invalid email format error.
On the newsletter form, submit an empty field, notanemail, and a@b and all should be rejected with validation errors.
On the contact form, submit <script>alert('xss')</script> in the name or message field. No alert dialog should fire.
On the contact form, submit `">` in the name or message field. No HTML should render or execute.
Submit valid inputs on both forms and they should pass validation and reach the API successfully.

after 3 approvals

### DIFF
--- a/components/NewsletterSubscribe/NewsletterForm/index.js
+++ b/components/NewsletterSubscribe/NewsletterForm/index.js
@@ -71,6 +71,12 @@ const NewsletterForm = ({ getReCaptchaToken }) => {
       return null;
     }
 
+    const emailRegex = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
+    if (!emailRegex.test(email)) {
+      setError('Please enter a valid email address');
+      return null;
+    }
+
     const confirmValidateRecaptcha = await validateReCaptcha();
     if (confirmValidateRecaptcha) {
       setName('');
--- a/components/NewsletterSubscribe/NewsletterForm/index.js
+++ b/components/NewsletterSubscribe/NewsletterForm/index.js
@@ -4,6 +4,7 @@ import { decode } from 'html-entities';
 import { NewsLetterSubmitButton } from '@/components/buttons/SubmitButton';
 import styles from './NewsletterForm.module.scss';
 import Container from '@/components/containers/Container';
+import { newsletterSchema } from '@/utils/schemas/newsletter';
 
 const NewsletterForm = ({ getReCaptchaToken }) => {
   const [error, setError] = useState(null);
@@ -61,19 +62,9 @@ const NewsletterForm = ({ getReCaptchaToken }) => {
 
     setError(null);
 
-    if (!name) {
-      setError('Please enter a name');
-      return null;
-    }
-
-    if (!email) {
-      setError('Please enter a valid email address');
-      return null;
-    }
-
-    const emailRegex = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
-    if (!emailRegex.test(email)) {
-      setError('Please enter a valid email address');
+    const result = newsletterSchema.safeParse({ name, email });
+    if (!result.success) {
+      setError(result.error.issues[0].message);
       return null;
     }
 
--- a/components/NewsletterSubscribe/NewsletterForm/index.js
+++ b/components/NewsletterSubscribe/NewsletterForm/index.js
@@ -153,18 +153,12 @@ const NewsletterForm = ({ getReCaptchaToken }) => {
               <div className={styles.formSending}>Sending...</div>
             )}
             {status === 'error' || error ? (
-              <div
-                className={styles.formError}
-                dangerouslySetInnerHTML={{
-                  __html: error || getMessage(message),
-                }}
-              />
+              <div className={styles.formError}>
+                {error || getMessage(message)}
+              </div>
             ) : null}
             {status === 'success' && status !== 'error' && !error && (
-              <div
-                className={styles.formSuccess}
-                dangerouslySetInnerHTML={{ __html: decode(message) }}
-              />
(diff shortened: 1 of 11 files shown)


components/NewsletterSubscribe/NewsletterForm/index.js
```diff
@@ -153,18 +159,12 @@ const NewsletterForm = ({ getReCaptchaToken }) => {
...
+                {error || getMessage(message)}
+              </div>
             ) : null}
             {status === 'success' && status !== 'error' && !error && (
-              <div
-                className={styles.formSuccess}
-                dangerouslySetInnerHTML={{ __html: decode(message) }}
-              />
+              <div className={styles.formSuccess}>{decode(message)}</div>
             )}
```
Satoshi-Sh, line 160: It's not directly related to your update, but the condition on lines 161-168 looks a bit off. Would this work?

```
{(status === 'error' || error) && (
  <div className={styles.formError}>
    {error || getMessage(message)}
  </div>
)}

{status === 'success' && !error && (
  <div className={styles.formSuccess}>
    {decode(message)}
  </div>
)}
```

shayla-develops-webs:
@Satoshi-Sh Good call out. I went ahead and updated.

pages/api/validateReCaptcha.js
```diff
@@ -19,6 +19,28 @@ export default async function handler(req, res) {
       });
     }
 
+    // Validate email format
+    const emailRegex = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
```
mtkksk1780: How about setting `emailRegex` in `utils/config.js` and importing it from there, since it’s used across multiple files below?

`components/NewsletterSubscribe/NewsletterForm/index.js`
`pages/api/validateReCaptcha.js`

mtkksk1780 (approved):
@shayla-develops-webs 
Thanks for the thorough investigation and improvements! It must have taken quite a bit of time.
I have left one small suggestion, which I hope will be helpful for further updates.

pages/api/validateReCaptcha.js
cherylli: or something like zod? it works well with react hook forms

shayla-develops-webs: @cherylli @mtkksk1780 I haven't worked with Zod before, but this is a great excuse to dive in! I'll look into it and see how it fits with the existing setup.

cherylli: have a look and see what you think. A proper email regex is very complicated, thats why I suggeest using existing validation packages like zod. Joi us another popular one

react hook forms + zod is super popular you should be able to find a lot of examples

shayla-develops-webs:
Ok so currently, the newsletter form validates manually with Zod, while the contact form uses react-hook-form's built-in Zod resolver. The form only has 2 fields so the manual approach works fine but should we refactor the newsletter form to to a react-hook-form for consistency across all forms, or is the split acceptable?

pages/api/validateReCaptcha.js
```diff
@@ -19,6 +19,28 @@ export default async function handler(req, res) {
       });
     }
 
+    // Validate email format
```
cherylli: I think we could remove all these validations now we have a zod validation schema which validates the form input
If we want to have validation here as well, we should probably use(import) the same validation schema defined in utils/ to be more consistent

utils/schemas/contact.js
```diff
@@ -0,0 +1,11 @@
+import { z } from 'zod';
+import { newsletterSchema } from './newsletter';
+
+export const contactSchema = newsletterSchema.extend({
```
Satoshi-Sh, line 4: My personal choice. 

I feel better to use z.object instead of newsletterSchema.extend as we can see all the fields in one file.

components/NewsletterSubscribe/NewsletterForm/index.js
```diff
@@ -4,6 +4,7 @@ import { decode } from 'html-entities';
 import { NewsLetterSubmitButton } from '@/components/buttons/SubmitButton';
 import styles from './NewsletterForm.module.scss';
 import Container from '@/components/containers/Container';
+import { newsletterSchema } from '@/utils/schemas/newsletter';
 
 const NewsletterForm = ({ getReCaptchaToken }) => {
   const [error, setError] = useState(null);
```
Satoshi-Sh, line 10: Nice to have: consider using RHF's register/reset here like ContactUsForm does, but out of scope for this PR — approving as-is.

Satoshi-Sh (approved):
Looks good to me. Thanks for your work!!

cherylli (approved):
thanks for the updates! Sorry didn't get a notification for the updates last week

Issue #312, opened 2026-07-12
Create Why Choose Us section


mariana-caldas:
## What do we need to build?

Implement the **Why Choose Us** section based on the approved Services Page mockups.

This section highlights Web Dev Path's core value proposition using a heading, supporting copy, and three informational cards that summarize the team's approach and expertise.

The implementation should follow the approved design and reuse existing Design System components whenever possible.

---

## SDD References

Before starting implementation, review the following documents:

- `specs/services-page/specification.md`
- `specs/services-page/decisions.md`
- `specs/services-page/references.md`

These documents are the source of truth for this initiative.

The approved Figma mockups define the intended layout and visual design, while the content documentation referenced in `references.md` contains the finalized copy for this section.

---

### Section Layout

Implement the **Why Choose Us** section according to the approved Figma mockups, including:

- Section label.
- Main heading.
- Supporting paragraph.
- Three value cards.
- Responsive desktop and mobile layouts.

---

### Value Cards

Implement the three informational cards using reusable components.

The cards are:

- Client-Centric
- Tailored Solutions
- End-to-End Expertise

Each card includes:

- Illustration
- Title
- Supporting description

---

## Implementation Notes

There are no known design or product dependencies for this section.

The implementation should follow the approved mockups and existing Design System.

---

## Approach suggestions

- Reuse existing card components from the Design System whenever possible.
- Keep the card component reusable for future sections.
- Separate content from presentation to simplify future updates.
- Maintain consistent spacing, typography, and responsive behavior with the rest of the Services page.

---

## Deadline

Please keep in mind that once you assign this task to yourself, it should be completed within **10 days**.

---

### Implementation

- [ ] Section matches the approved Figma mockups.
- [ ] Three value cards are implemented.
- [ ] Illustrations, titles, and descriptions match the approved content.
- [ ] Desktop and mobile layouts are responsive.
- [ ] Components are reusable where practical.

### Validation

- [ ] Review implementation against `specification.md`.
- [ ] Review implementation against `decisions.md`.
- [ ] Review implementation against the approved Figma mockups.
- [ ] Verify responsive behavior.
- [ ] Verify browser compatibility (Chrome, Firefox, Edge, and Safari).

- Test the section across different screen sizes using the browser's inspection tools.
- Verify that changes do not negatively affect existing shared components or layouts.
- If build issues occur before submitting the Pull Request, run `yarn build` locally, resolve any errors, and commit the fixes.
[...]

PR #315, opened 2026-07-18
SDD Services dev -> Main

mariana-caldas:

Yes

#### What is this change?

We're adding the SDD documentation from `services-dev` into `main` if the team agrees this is a good time to do that.

After review, and if we decide it's a good idea to send that approach to main already. 
I'd vote "yes" because it gives visibility for what we're building when people visit our repository. You can also start adding that to your resumes/LinkedIn


cherylli:
> We're adding the SDD documentation from services-dev into main if the team agrees this is a good time to do that.

I'm not sure, I thought the idea of commiting to the service branch is to wait till it's ready (all done) to merge into main. However, this is only documentation, so it's probably fine.

>  it gives visibility for what we're building when people visit our repository

I think they can see from the Issues tab

----
If we want to do it we need to do it before there are service related code changes. Sorry, can't believe I missed this PR, even though I have checked the repo a few times since then...

shayla-develops-webs (approved):
Looks great! Would like to have this merged in before making any code changes to services-dev.

mariana-caldas:
We don't need to send the SDD references to `main` now, @cherylli  and @shayla-develops-webs , so I'm closing this PR.
One thing to keep in mind, though: issues need to include an SDD section instead of adding resource links directly, as we have here: https://github.com/Web-Dev-Path/web-dev-path/issues/312. That way, we help devs get used to referring to the `services` folder for information.

